PR #253, opened 2014-11-05
Access FieldData as a Runtime service

cpennington:


xblock/core.py
```diff
@@ -1,26 +1,33 @@
...
+from xblock.exceptions import DisallowedFileError
+from xblock.fields import String, List, Scope
+from xblock.mixins import ScopedStorageMixin, HierarchyMixin, RuntimeServicesMixin, HandlersMixin
 from xblock.plugin import Plugin
 from xblock.validation import Validation
 
 
 # __all__ controls what classes end up in the docs.
 __all__ = ['XBlock']
+UNSET = object()
```
kdmccormick, line 19: WHY

nedbat: So that any value passed will be recognized, including None.  `object()` gives us a unique object that will fail `x is UNSET` for any possible value passed in as x.   Unless I'm misunderstanding your question?

nedbat: Yeah, I can see that. I don't remember why we were being so cautious.

PR #277, opened 2014-11-21
CourseService

rmachielse:
Adding CourseService to allow XBlocks to access the current course data

This pull request allows XBlocks to access user and course credentials through a service.
Technical documentation can be found in the comments on edx/edx-platform#6030


nedbat:
We are not making progress on this course information service, because it seems like no XBlock author needs it at the moment.  As much as I hate to close pull requests that have useful code in them, I'm going to close this for now.  We can re-open it when we come back to the work.

PR #285, opened 2015-01-07
Initial xblock-javascript page

mhoeber:


mhoeber:
@cpennington @doctoryes @dmitchell 

I have no doubt this needs more work to be useful, but I didn't want to keep putting off creating this PR.

Some specific questions:
- Are there good examples of XBlocks that are using children that I can show (didn't see or missed in the SDK examples)
- When/why would an XBlock use children vs. child map from the runtime.

Thanks

doc/guide/javascript.rst
```diff
@@ -0,0 +1,234 @@
...
+    . . .
+
+    children: function(block) {
+          return $(block).prop('xblock_children');
+        },
+    . . . 
+
+NEED EXAMPLE OF XBLOCK USING THIS
+
+WHEN XBLOCK USES THIS VS. MAP
```
cpennington: An XBlock would use this to iterate over it's children (when order matters).

cpennington: For instance, the Sequence module does this, I believe.

doc/guide/javascript.rst
```diff
@@ -0,0 +1,234 @@
...
+      for (var i = 0; i < children.length; i++) {
+        var child = children[i];
+        if (child.name == childName) {
+          return child
+        }
+      }
+    }
+    . . . 
+
+NEED EXAMPLE OF XBLOCK USING THIS
```
cpennington: An XBlock uses this when it needs to perform different actions on different children. The only example of this that I'm aware of is the `problem_demo` block in the XBlock SDK.

doc/guide/javascript.rst
```diff
@@ -0,0 +1,234 @@
...
+
+You must add the JavaScript file to the fragment in your XBlock.  For example,
+`thumbs.py`_ loads the JavaScript above into the fragment for the student view
+of the Thumbs XBlock:
+
+.. code-block:: Python
+  
+  js_str = pkg_resources.resource_string(__name__, "static/js/src/thumbs.js")
+  frag.add_javascript(unicode(js_str))
+  frag.initialize_js('ThumbsBlock')
```
cpennington: You should probably call out the separate actions of `add_javascript` (which gets the javascript onto the page), and `initialize_js`, which specifies which initialization function to call to initialize the XBlock (this initialization function should be provided by the javascript added in `add_javascript`).

mhoeber:
@cpennington   I updated per your comments.  Do you think we should merge as is (and create follow up stories as needed) or would you add anything missing now?

doc/guide/javascript.rst
```diff
@@ -0,0 +1,267 @@
...
+.. code-block:: JavaScript
+
+    . . .
+
+    children: function(block) {
+          return $(block).prop('xblock_children');
+        },
+    . . . 
+
+An XBlock uses the children function when it needs to iterative over an
```
cpennington: s/iterative/iterate/

cpennington:
Minor wording change. otherwise, :+1:

mhoeber:
@cpennington   Travis builds were passing until I made last fix. Errors not related to this change.  OK to merge?

cpennington:
Yes

PR #286, opened 2015-01-08
index_view exposable in XBlocks

martynjames:
@cdodge, @chrisndodge & maybe @mattdrayer  - Please review and then I'll open up to a wider. This is the base implementation of the IndexInfoMixin to be used for XBlocks to override when providing index information.

### DIFF
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -458,3 +458,18 @@ class XmlSerializationMixin(ScopedStorageMixin):
             node.insert(0, elem)
         else:
             node.set(field_name, unicode(field.read_from(self)))
+
+
+class IndexInfoMixin(object):
+    """
+    This mixin provides interface for classes that wish to provide index
+    information which might be used within a search index
+    """
+
+    def index_dictionary(self):
+        """
+        return key/value fields to feed an index within in a Python dict object
+        values may be numeric / string or dict
+        default implementation is an empty dict
+        """
+        return {}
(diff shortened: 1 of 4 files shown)


chrisndodge:
@martynjames looks about right to me

martynjames:
@cpennington - is this okay with you? Preparation for the indexing of xblocks for searchability

xblock/mixins.py
```diff
@@ -447,3 +447,18 @@ def _add_field(self, node, field_name, field):
...
             node.set(field_name, unicode(field.read_from(self)))
+
+
+class IndexInfoMixin(object):
+    """
+    This mixin provides interface for classes that wish to provide index
+    information which might be used within a search index
+    """
+
+    def index_view(self):
```
cpennington: I would probably call this something other than `_view`, since it doesn't return a `Fragment` (unlike other XBlock views)p

martynjames: Thanks @cpennington - @cdodge had suggested `index_view`. I had originally had `index_info` but that seems very generic... how about `index_dictionary`?

martynjames: @cpennington, @chrisndodge, @cdodge - Hoping to get resolution on this one (and hopefully then 2 thumbs up???). Anyone got any objections to calling this method ```index_dictionary``` instead of ```index_view```?

chrisndodge: sounds good to me :-)

cpennington: works for me as well.

cpennington:
:

cpennington:
:+1:

PR #288, opened 2015-01-28
Ignore comments when parsing XML

bradenmacdonald:
**Description**: This is a simple change to ignore comments when parsing XML *using the default parser built into XBlock*. Currently, if the XML includes comments, an error will result and the XBlock tree construction fails. For example, right now in general you cannot put comments into xblock-sdk workbench scenarios.

**Use cases**:
* Adding explanatory comments to course exports or workbench scenarios
* Temporarily disabling blocks in XML by commenting them out
* Some XBlocks currently have an "editor" interface that's just a big XML field where users can edit the XML that defines them and their child blocks. We currently have to override the `parse_xml` method in each such block to add code to ignore comments in order to avoid errors when users add them as part of an edit.

**Affects**:
* Any XBlock that uses the default XML parser to load itself and/or its children from an XML representation. (e.g. during course import and export, in the XBlock SDK workbench)

**Does not affect**:
* Any XBlocks that implement their own `parse_xml` method
* Blocks that store XML as a string in one of their fields.
* CAPA problem XML - capa XML is stored as a string in the "data" field, has its own parser, and is not built of XBlock XML but rather a separate XML language specific to capa.

**What about serialization**? This ignores the comments, so if you create an XBlock from XML, then ask that XBlock to export itself as XML [using the default serialization code](https://github.com/edx/XBlock/blob/master/xblock/mixins.py#L400), the XML is re-created from scratch and the comments will be gone. Note that this does not break any existing behaviour as comments are not currently allowed unless specifically implemented by a particular XBlock. As well, this does not affect capa problems or blocks that store XML in fields other than "content" or that do their own XML parsing/export.

However, XBlocks that use XML to define their configuration/children and that wish to preserve comments can do so using a custom XML parser, as shown in one of the new test cases.

### DIFF
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -374,6 +374,8 @@ class XmlSerializationMixin(ScopedStorageMixin):
         # The base implementation: child nodes become child blocks.
         # Or fields, if they belong to the right namespace.
         for child in node:
+            if child.tag is etree.Comment:
+                continue
             qname = etree.QName(child)
             tag = qname.localname
             namespace = qname.namespace
--- a/xblock/runtime.py
+++ b/xblock/runtime.py
@@ -682,8 +682,7 @@ class Runtime(object):
 
     def parse_xml_file(self, fileobj, id_generator=None):
         """Parse an open XML file, returning a usage id."""
-        parser = etree.XMLParser(remove_comments=True)
-        root = etree.parse(fileobj, parser=parser).getroot()
+        root = etree.parse(fileobj).getroot()
         usage_id = self._usage_id_from_node(root, None, id_generator)
         return usage_id
 
--- a/xblock/runtime.py
+++ b/xblock/runtime.py
@@ -682,7 +682,8 @@ class Runtime(object):
 
     def parse_xml_file(self, fileobj, id_generator=None):
         """Parse an open XML file, returning a usage id."""
-        root = etree.parse(fileobj).getroot()
+        parser = etree.XMLParser(remove_comments=True)
+        root = etree.parse(fileobj, parser=parser).getroot()
         usage_id = self._usage_id_from_node(root, None, id_generator)
         return usage_id
 
--- a/xblock/test/test_parsing.py
+++ b/xblock/test/test_parsing.py
@@ -132,6 +132,20 @@ class ParsingTest(XmlTest, unittest.TestCase):
         self.assertEqual(child2.parent, block.scope_ids.usage_id)
 
     @XBlock.register_temp_plugin(Leaf)
+    @XBlock.register_temp_plugin(Container)
+    def test_xml_with_comments(self):
+        block = self.parse_xml_to_block("""\
+                    <!-- This is a comment -->
+                    <container>
+                        <leaf data1='child1'/>
+                        <!-- <leaf data1='ignore'/> -->
+                        <leaf data1='child2'/>
+                    </container>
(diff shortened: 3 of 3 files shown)


edx-webhook:
Thanks for the pull request, @bradenmacdonald! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?number=288&repo=edx%2FXBlock

edx-webhook:
Thanks for the pull request, @bradenmacdonald! I've created [OSPR-371](https://openedx.atlassian.net/browse/OSPR-371) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

We can't start reviewing your pull request until you've added yourself to the [AUTHORS](https://github.com/edx/XBlock/blob/master/AUTHORS) file. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

e-kolpakov:
@bradenmacdonald :+1:

sarina:
@carsongee could I have you take a look at this change, since your customers may still be authoring in XML?

carsongee:
I'm not super familiar with this, but would this mean that comments aren't actually stored in the module store?  I.e. I wouldn't be able to leave comments for myself in the problem and see them in "Staff Debug", or is this just during serialization for rendering and such?

e-kolpakov:
@carsongee If I recall right - quite the opposite: XML is saved as is, but XML parser that deserializes XBlocks will ignore it (and not crash). @bradenmacdonald could you please confirm this?

bradenmacdonald:
@e-kolpakov is correct - currently, the XBlock parser actually throws an error if it encounters a comment. This change permits comments to be present in a place where they were previously not working.

@carsongee I assume you are talking about capa problems - they store their XML in a "data" field and they use a different parser [which ignores comments](https://github.com/edx/edx-platform/blob/master/common/lib/capa/capa/capa_problem.py#L679), so this change doesn't affect that in any way. Comments put into problem XML will be preserved as they currently are.

carsongee:
Cool, thanks for the information. This seems fine to me then.

sarina:
@bradenmacdonald could you update your description to be clearer about what component(s) this PR affects, so that people looking at this after-the-fact will have a better understanding about what is going on? I didn't realize that Capa problems used a different parser, for example. Please also explain if this preserves comments in serialized XML.

bradenmacdonald:
@sarina Good point. I have updated this PR description to hopefully be a lot more clear. I have also changed the approach I took with the code so that XBlocks can optionally override this new behaviour to do things like serialize comments (no comments will be serialized by default - see details in revised description). 

I also added a new test demonstrating how an XBlock can preserve comments in its XML if it wants to.

(Also to be clear, none of this affects capa problem XML in any way)

sarina:
Cool - this is an excellent change. (@carsongee sorry to tag you earlier, I thought this was in edx-platform not xblock repo) :+1:

PR #289, opened 2015-02-20
Backup class cache in register_temp_plugin

cpennington:


cpennington:
@nedbat @doctoryes: This is needed for https://github.com/edx/edx-platform/pull/7050

nedbat:
One wonders why PLUGIN_CACHE is not a class attribute, but this fixes the test isolation. :+1:

cpennington:
Because if it were, then we'd be back to the bad old days of having multiple caches depending on which of `XModule`, `XBlock`, or `XModuleDescriptor` you happened to try and load the plugins through.

doctoryes:
:+1:

nedbat:
We could make it Plugin.global_cache, and access it that way, not as cls.global_cache.

cpennington:
Yeah, I guess that's true. Better, though?

nedbat:
You already have my :+1: :)

PR #290, opened 2015-03-11
Clarify Apache 2.0 license

singingwolfboy:
The code in this repository used to be licensed under the AGPL license, but it was switched to the Apache 2.0 license in #231. Looks like we missed a few references, though. @nedbat, can you review this?


nedbat:
:+1: For bonus points, add a trailing comma in the setup() call.

PR #293, opened 2015-04-01
Fix links in getting_started.rst

clytwynec:
@nedbat @mhoeber please review

These links were to pages that have been moved from edx-platform to edx-documentation.  I've updated them to point to the right place.

### DIFF
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -141,14 +141,12 @@ The simplest test environment is the XBlock workbench. It can be found in the
 Testing with the edX LMS
 ........................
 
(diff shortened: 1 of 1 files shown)


nedbat:
Maybe it would be better to link to readthedocs, so they have the full navigation?  http://edx-developer-guide.readthedocs.org/en/latest/extending_platform/xblocks.html

clytwynec:
@nedbat  Good idea. Done.

mhoeber:
:+1:

nedbat:
:+1:

PR #295, opened 2015-04-10
EXPLICITLY_SET always overrides _dirty_fields content

e-kolpakov:
**Background and Motivation:** edx-platform import routine includes updating xblock location and explicitly [setting some fields to their values](https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/modulestore/xml_importer.py#L1160-1171) in order to mark them as dirty (note the comment). However, this found to be wrong, as MUTABLE fields (e.g. `List` and `Dict`) are marked as dirty [*on read*](https://github.com/edx/XBlock/blob/master/xblock/fields.py#L490-491). As a result, setting mutable field to the same value does not mark it as dirty, as ["baseline"](https://github.com/edx/XBlock/blob/master/xblock/fields.py#L419) was just set to the same value. This, in turn, breaks importing later, but only for drafts, as published versions does not pass through `_update_module_location` method that updates the location to draft. As a result, mutable fields e.g. `children`) are not saved with block.

So the issue was traced down to [`_mark_dirty`](https://github.com/edx/XBlock/blob/master/xblock/fields.py#L408) function - it looks like, despite being explicitly set when parsing xml, neither `children`, nor `correct_choices` (both of type `List`) were not marked as dirty. Hence the proposed fix - always mark fields as dirty if explicitly assigned.

Also, particular feature of this bug is that it affects only draft versions - published work just fine. Side effect - LMS shows correct data.

**JIRA Ticket**: [SOL-810](https://openedx.atlassian.net/browse/SOL-810)
**Related:** [OSPR-517](https://openedx.atlassian.net/browse/OSPR-517), [PR 299(closed)](https://github.com/edx/XBlock/pull/299) - contains discussion about why this PR was reactivated; [SOL-811](https://openedx.atlassian.net/browse/SOL-811)
**Sandbox**: [LMS](http://sandbox3.opencraft.com/), [Studio](http://sandbox3.opencraft.com:18010/)
**Partner information:** 3rd party-hosted open edX instance, for an edX solutions client.
**Merge dealine:** preferably Friday, May 1st (in time for next release).
**Scope:** Probably any kinds of XBlocks that have children and `Scope.settings` fields simultaneously. Might not be the case for non-split modulestores.
**Test Instructions:** No user-facing changes, use test instructions for [edx-platform import fix PR](https://github.com/edx/edx-platform/pull/7816)

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -405,7 +405,7 @@ class Field(Nameable):
 
         # Deep copy the value being marked as dirty, so that there
         # is a baseline to check against when saving later
-        if self not in xblock._dirty_fields:
+        if self not in xblock._dirty_fields or value is EXPLICITLY_SET:
             xblock._dirty_fields[self] = copy.deepcopy(value)
 
     def _is_dirty(self, xblock):
--- a/xblock/test/test_fields.py
+++ b/xblock/test/test_fields.py
@@ -26,7 +26,9 @@ from xblock.fields import (
     UNIQUE_ID
 )
 
-from xblock.test.tools import assert_equals, assert_not_equals, assert_not_in, TestRuntime
+from xblock.test.tools import (
+    assert_equals, assert_not_equals, assert_in, assert_not_in, assert_false, assert_true, TestRuntime
+)
 from xblock.fields import scope_key, ScopeIds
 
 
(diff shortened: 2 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2FXBlock&number=295

edx-webhook:
Thanks for the pull request, @e-kolpakov! I've created [OSPR-517](https://openedx.atlassian.net/browse/OSPR-517) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

e-kolpakov:
@sarina this PR contains fix for another issue we found when working with mentoring, namely, export-import does not preserve children and scope settings. Code changes are really tiny, but the scope of impact might be big - specifically it might cause lot more storage operations to happen when saving XBlocks. Since it's not clear if it was a bug or an intentionally put behavior, it would be best if someone from core XBlock team reviewed this.

Kelketek:
@e-kolpakov I'm a little confused on what I'm seeing in your directions and your explination of the bug. Your explanation says this bug only happens for drafts, and not for published work. Your directions tell me to publish to verify the bug-- that seems like a contradiction.

I have applied the patch, and after having done so, published import/export works, but drafts do not. An `ItemNotFoundError` shows up during import, and the children are missing.

e-kolpakov:
@Kelketek publishing occurs before exporting-importing and frankly, is optional. Basically steps 0-9 are preparation steps. Could you please elaborate further on that `ItemNotFoundError` - I haven't seen it when working on this.

Kelketek:
@e-kolpakov http://pastebin.com/MrUycKeg for the traceback. Here's how to reproduce:

1. Create a new unit with a new Mentoring block (don't use an existing one) and fill it with questions and whatnot as you described. Do not publish it.
2. Export.
3. Import to another course.

You will find that the children are missing. It should be noted that if you have a published version of the Mentoring block already in place, it will just discard the draft entirely and revert to the last published version.

sarina:
@e-kolpakov : sorry for the delay, I'm still catching up from PyCon.

I don't feel qualified to review this pr. I've asked @nedbat to take a look, and he'll ping anyone else who may need to take a look as well.

cpennington:
@e-kolpakov: If you restart studio in between steps 10 and 11, does the problem still happen? If it does, then we can shorten the testing instructions to "import this known bad XML". If it doesn't, then it seems likely that the problem is related to caching of XBlock objects in one of the modulestores.

e-kolpakov:
@cpennington actually, it seems the issue is a bit more complicated as this fix does not fix everything: we observe the same behavior (missing children) when importing a course that was exported when it had no published versions (I mean, nothing was ever published manually). Also, it seems like importing such a course results in published version being created, i.e. what was in draft stays in draft, but there's also published version with the same content, and it's visible in studio.

Looks like the problem with import-export is a bit deeper. I'll do what you suggested (probably today, depends on how other stuff goes) and update the description.

e-kolpakov:
@cpennington I've updated PR description with my latest findings. Any suggestions what might be causing such behavior?

cpennington:
@e-kolpakov: If this is all about strangeness during import/export, I would check with @doctoryes, who is working on some SplitModulestore import/export troubles. Or, try using the old-mongo modulestore, to see if the behavior is different (if it is, that will help with debugging).

Also, I would tend to suspect that import/export problems are edx-platform specific, until/unless you can reproduce it in xblock-sdk using a workbench scenario, or reproduce the specific issues in an xblock library test. It would be helpful to isolate the behavior you expect as a unit test in the XBlock library that just sets and reads fields, if that's where you think the underlying behavior issue is.

e-kolpakov:
The issues seems to be in edx-platform import-export code, so this change only masks the problem - closing the PR.

cpennington:
:+1:

cpennington:
@doctoryes, will you take a look at this as well, so that we can get it merged?

doctoryes:
:+1:

PR #297, opened 2015-04-15
existent has an e, not an a

nedbat:
@sarina I think I have this right?


sarina:
Yeah. Did I write this? seems like my type of spelling.

:+1:

nedbat:
I didn't look to see who wrote it, just figured you'd be good for good eye and a quick thumb.

PR #298, opened 2015-04-23
Declaring `allow_duplicate = True` in the keyword argument in field defi...

awaisdar001:
Added a keyword argument `allow_duplicate` in the field definition which will identify which fields can be  duplicated.

[TNL-1754] (https://openedx.atlassian.net/browse/TNL-1754)


awaisdar001:
@adampalay @cpennington Please review

cpennington:
This shouldn't be a change to XBlock core. edx-platform should just use `.runtime_options['allow_duplicate']`, which is would be set by passing a keyword argument through `**kwargs`.

awaisdar001:
@cpennington Ok, cool. 
using this way, i was able to check `field.allow_duplicate == True`

so if we skip this change in the core xblock, we would need to check that through `field.runtime_options` 
Right?

PR #301, opened 2015-06-18
Adam/add force save fields 2

adampalay:
@cpennington , not ready for review yet, but I'm running into an issue with this test: https://github.com/edx/XBlock/blob/master/xblock/test/test_core.py#L643

When I pdb through that test, I notice that `mutable_test.list_field` is in the list of dirty_fields, but returns false when `_is_dirty` is run on it.

Any insights into what I'm missing here?

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -498,11 +498,22 @@ class Field(Nameable):
         Setting a value does not update the underlying data store; the
         new value is kept in the cache and the xblock is marked as
         dirty until `save` is explicitly called.
+
+        Note, if there's already a cached value and it's equal to the value
+        we're trying to cache, we won't do anything.
         """
         value = self._check_or_enforce_type(value)
-        # Mark the field as dirty and update the cache:
-        self._mark_dirty(xblock, EXPLICITLY_SET)
-        self._set_cached_value(xblock, value)
+        cached_value = self._get_cached_value(xblock)
+        try:
+            value_has_changed = cached_value != value
+        except Exception:  # pylint: disable=broad-except
+            # if we can't compare the values for whatever reason
+            # (i.e. timezone aware and unaware datetimes), just reset the value.
+            value_has_changed = True
+        if value_has_changed:
+            # Mark the field as dirty and update the cache
+            self._mark_dirty(xblock, EXPLICITLY_SET)
+            self._set_cached_value(xblock, value)
 
     def __delete__(self, xblock):
         """
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -405,7 +405,7 @@ class Field(Nameable):
 
         # Deep copy the value being marked as dirty, so that there
         # is a baseline to check against when saving later
-        if self not in xblock._dirty_fields or value is EXPLICITLY_SET:
+        if self not in xblock._dirty_fields:
             xblock._dirty_fields[self] = copy.deepcopy(value)
 
     def _is_dirty(self, xblock):
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -244,32 +244,54 @@ class ScopedStorageMixin(RuntimeServicesMixin):
         if not self._dirty_fields:
             # nop if _dirty_fields attribute is empty
             return
-        try:
-            fields_to_save = self._get_fields_to_save()
-            # Throws KeyValueMultiSaveError if things go wrong
-            self._field_data.set_many(self, fields_to_save)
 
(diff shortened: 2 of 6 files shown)


cpennington:
it's only `_is_dirty` if the value has changed. `dirty_fields` is just the list of fields that might be dirty.

xblock/fields.py
```diff
@@ -498,11 +498,15 @@ def __set__(self, xblock, value):
...
         dirty until `save` is explicitly called.
+
+        Note, if there's already a cached value and it's equal to the value
+        we're trying to cache, we won't do anything.
         """
         value = self._check_or_enforce_type(value)
-        # Mark the field as dirty and update the cache:
-        self._mark_dirty(xblock, EXPLICITLY_SET)
-        self._set_cached_value(xblock, value)
+        if self._get_cached_value(xblock) != value:
```
adampalay: @cpennington , in the tests, I've run into scenarios where this line fails because the value of the field stored on the xblock can't be compared to the value we're trying to set on it. For example, if `self._get_cached_value(xblock)` is a non-timezone-aware `datetime` object, and `value` _is_ a timezone-aware `datetime` object, then trying to compare them throws a `TypeError`. Is this an error condition I should be anticipating and catching? I assume we only have timezone-aware datetimes in production, but it's possible in other scenarios we try to set a field to something it can't currently be compared to.

adampalay:
@cpennington , @waheedahmed , review please?

xblock/fields.py
```diff
@@ -498,11 +498,22 @@ def __set__(self, xblock, value):
...
+        cached_value = self._get_cached_value(xblock)
+        try:
+            value_has_changed = cached_value != value
+        except Exception:  # pylint: disable=broad-except
+            # if we can't compare the values for whatever reason
+            # (i.e. timezone aware and unaware datetimes), just reset the value.
+            value_has_changed = True
+        if value_has_changed:
+            # Mark the field as dirty and update the cache
+            self._mark_dirty(xblock, EXPLICITLY_SET)
```
cpennington: I wonder, for a mutable field, whether this should just do the deepcopy of value. What happens if we do (on a mutable field): set -> save -> mutate -> save?

adampalay: @cpennington I just added a test for that sequence, which passes. Let me know if I wrote the test correctly

xblock/mixins.py
```diff
@@ -244,32 +245,48 @@ def save(self):
...
-        fields_to_save = {}
+        fields_to_save = []
         for field in self._dirty_fields.keys():
             # If the field value isn't the same as the baseline we recorded
             # when it was read, then save it
             if field._is_dirty(self):  # pylint: disable=protected-access
-                fields_to_save[field.name] = field.to_json(self._field_data_cache[field.name])
+                fields_to_save.append(field.name)
+
         return fields_to_save
```
cpennington: Probably doesn't need to happen with this PR, but this could be simplified and inlined as:

    [field.name for field in self._dirty_fields if field.is_dirty(self)]

xblock/test/test_fields_api.py
```diff
@@ -163,12 +163,26 @@ def test_delete_with_save_writes(self):
...
+    def test_set_after_get_doesnt_save(self):
+        self.set(self.get())
+        # pylint: disable=protected-access
+        fields_to_save = self.block._get_fields_to_save()
+        assert_equals(len(fields_to_save), 0)
+
+        self.set(self.new_value)
+        # pylint: disable=protected-access
+        fields_to_save = self.block._get_fields_to_save()
+        assert_equals(len(fields_to_save), 1)
```
cpennington: I think this test would be better written to test that `patched_set_many` is or isn't called after each call to `save`.

xblock/test/test_fields_api.py
```diff
@@ -206,6 +220,23 @@ def test_mutation_without_save_makes_non_default(self):
         self.mutate(self.get())
         assert_false(self.is_default())
 
+    def test_mutate_pointer_after_save(self):
+        pointer = self.get()
+        self.mutate(pointer)
+        self.block.save()
```
cpennington: Maybe add an `assert_equals` here too? And perhaps a comment on the next line that we're testing what happens if you mutate a pointer without retrieving it through `.get`?

adampalay:
@cpennington , responded to comments. Ok now?
@waheedahmed ?

cpennington:
+1

waheedahmed:
:+1:

PR #302, opened 2015-06-19
Pass parent to xblock on load

cpennington:
@doctoryes @alawibaba @utkjad: This should help optimize parent traversal.

### DIFF
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -332,18 +332,58 @@ class HierarchyMixin(ScopedStorageMixin):
         # A cache of the parent block, retrieved from .parent
         self._parent_block = None
         self._parent_block_id = None
+        self._child_cache = {}
+
+        for_parent = kwargs.pop('for_parent', None)
+
+        if for_parent is not None:
+            self._parent_block = for_parent
+            self._parent_block_id = for_parent.scope_ids.usage_id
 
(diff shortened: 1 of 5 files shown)


xblock/mixins.py
```diff
@@ -345,6 +352,33 @@ def get_parent(self):
             self._parent_block_id = self.parent
         return self._parent_block
 
+    @property
+    def has_cached_parent(self):
+        """Return whether this block has a cached parent block."""
+        return self.parent and self._parent_block is not None and self._parent_block_id == self.parent
```
doctoryes: Why the middle `is not None`?

xblock/runtime.py
```diff
@@ -628,14 +628,15 @@ def construct_xblock_from_class(self, cls, scope_ids, field_data=None, *args, **
         Construct a new xblock of type cls, mixing in the mixins
         defined for this application.
         """
-        return self.mixologist.mix(cls)(
+        block = self.mixologist.mix(cls)(
```
doctoryes: Was there a subtle reason behind this change?

cpennington: Nope, that was debugging detritus

doctoryes:
:+1:

alawibaba:
:+1:

PR #303, opened 2015-07-07
Xiaolu/shared field

beardeer:
@pmitros @cpennington @DongwookYoon 

Hi team,

This is another structure review request, this time it is on the XBlock end. Besides the implementation itself, I also added some test cases to ensure I am getting the right objects. 

Please let me know if you feel my code is sensible, thanks!

### DIFF
--- a/xblock/field_data.py
+++ b/xblock/field_data.py
@@ -11,6 +11,8 @@ from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 
 from xblock.exceptions import InvalidScopeError
+from xblock.core import XBlock
+from xblock.fields import ScopeIds
 
 
 class FieldData(object):
@@ -155,7 +157,10 @@ class SplitFieldData(FieldData):
         return self._scope_mappings[scope]
 
     def get(self, block, name):
-        return self._field_data(block, name).get(block, name)
+        if isinstance(block, XBlock):
+            return self._field_data(block, name).get(block, name)
+        elif isinstance(block, ScopeIds):
+            pass
 
     def set(self, block, name, value):
         self._field_data(block, name).set(block, name, value)
--- a/xblock/field_data.py
+++ b/xblock/field_data.py
@@ -220,11 +220,3 @@ class ReadOnlyFieldData(FieldData):
 
     def __repr__(self):
         return "ReadOnlyFieldData({!r})".format(self._source)
-
-class QueryableFieldData(SplitFieldData):
-    """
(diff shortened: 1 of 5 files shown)


xblock/query.py
```diff
@@ -0,0 +1,118 @@
...
+    def scope(self):
+        return self._remote_scope
+
+    @property
+    def xblock(self):
+        return self._xblock
+
+    @property
+    def name(self):
+        return self._name
```
cpennington: Do you need to make all of these read-only?

xblock/field_data.py
```diff
@@ -154,8 +156,24 @@ def _field_data(self, block, name):
...
+            TYPE: Description
+        """
+        return self._scope_mappings[remote_field_scope]
+
+
+    def get(self, block, name, remote_field_scope=None):
+        if isinstance(block, XBlock):
+            return self._field_data(block, name).get(block, name)
+        elif isinstance(block, ScopeIds):
+            return self._shared_field_data(remote_field_scope).get(block, name)
```
cpennington: I'm a bit confused. This looks like it's trying to make `get` work with either an `XBlock` or the `ScopeIds` from an `XBlock`, but then the behavior is different in the different cases (rather than just pulling the `ScopeIds` out of the `XBlock` in the former case). Why?

beardeer: I need the ScopeIds so that I can query on user id, while xblock is still necessary because I need to access the attaching field data. But let me review the workflow to ensure it is working as expected. 

Thank you!

xblock/query.py
beardeer: Although that is my intention but I will re-think possible use cases.

xblock/fields.py
```diff
@@ -146,6 +148,24 @@ def scopes(cls):
         return [cls.NONE, cls.ONE, cls.ALL]
 
 
+class SharingUserScope(object):
```
pmitros: Perhaps SharedUserScope?

pmitros:
It would be helpful to have an example XBlock showing how these are used, and possibly test cases around that. Right now, it's a little hard to wrap my mind around how all of this interacts without something concrete. 

I usually work the other way with API design -- I start with as concrete use case, figure out how to make it easy to write, and then develop the API behind it. A more extreme of this approach is TDD, which edX prefers.

beardeer:
Working on a demo XBlock that adopts my changes, closing this PR now.

PR #304, opened 2015-07-09
Add "supports" functionality tagging on xBlock views.

nasthagiri:
@cpennington @ormsbee Please review.

### DIFF
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -121,18 +121,20 @@ class RuntimeServicesMixin(object):
         super(RuntimeServicesMixin, self).__init__(**kwargs)
 
     @classmethod
-    def needs(cls, service_name):
-        """A class decorator to indicate that an XBlock class needs a particular service."""
+    def needs(cls, *service_names):
+        """A class decorator to indicate that an XBlock class needs particular services."""
         def _decorator(cls_):                                # pylint: disable=missing-docstring
-            cls_._services_requested[service_name] = "need"  # pylint: disable=protected-access
+            for service_name in service_names:
+                cls_._services_requested[service_name] = "need"  # pylint: disable=protected-access
             return cls_
         return _decorator
 
     @classmethod
-    def wants(cls, service_name):
-        """A class decorator to indicate that an XBlock class wants a particular service."""
+    def wants(cls, *service_names):
+        """A class decorator to indicate that an XBlock class wants particular services."""
         def _decorator(cls_):                                # pylint: disable=missing-docstring
-            cls_._services_requested[service_name] = "want"  # pylint: disable=protected-access
+            for service_name in service_names:
+                cls_._services_requested[service_name] = "want"  # pylint: disable=protected-access
             return cls_
         return _decorator
 
@@ -545,13 +547,13 @@ class ViewsMixin(object):
     This mixin provides decorators that can be used on xBlock view methods.
     """
     @classmethod
-    def supports(cls, functionality):
+    def supports(cls, *functionalities):
         """
         A view decorator to indicate that an xBlock view has support for the
-        given functionality.
+        given functionalities.
 
         Arguments:
-            functionality (string): An identifier for the functionality of the view.
+            functionalities: String identifiers for the functionalities of the view.
                 For example: "multi_device".
         """
         def _decorator(view):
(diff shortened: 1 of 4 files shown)


nasthagiri:
please review.

xblock/mixins.py
```diff
@@ -538,3 +538,50 @@ def index_dictionary(self):
...
         """
         return {}
+
+
+class ViewsMixin(object):
+    """
+    This mixin provides decorators that can be used on xBlock view methods.
+    """
+    @classmethod
+    def supports(cls, functionality):
```
cpennington: Nit: should this be `*functionalities`, so that you can do something like `@supports('multi_device', 'screen_reader', 'lynx')`?

cpennington:
Have you looked in the XBlock documentation to see if there's anywhere that this should be mentioned? @mhoeber: should this make its way into the XBlock tutorial?

cpennington:
Or, alternately, should `multi_device`

xblock/mixins.py
nasthagiri: Nice idea.  I'd like to be consistent with the other xBlock decorators.  It looks like `@needs` and `@wants` take only a single service name.  Should we update those as well if we like the list of strings approach?

nasthagiri:
@cpennington @mhoeber regarding documentation, we will need to work with product to determine when's the right time to educate the open source community on our xBlock tagging scheme.  @lou-wang Any thoughts on this?

mhoeber:
@cpennington @nasthagiri   I'm backlogging this story and will follow up.

xblock/mixins.py
cpennington: Yeah, it would be nice to keep them consistent. If you feel like making the change as part of this PR, that would be awesome, otherwise we can just leave them all as single-argument for now.

nasthagiri:
@cpennington I have added a commit to this PR to add multi-arg support for the xBlock decorators.  Please review and let me know if there's anything blocking this PR.  Thanks.

xblock/test/test_mixins.py
```diff
@@ -133,3 +133,78 @@ def test_index_info(self):
...
+                ("an_unsupported_view", "a_functionality", False),
+
+                ("multi_featured_view", "functionality1", True),
+                ("multi_featured_view", "functionality2", True),
+                ("multi_featured_view", "bogus_functionality", False),
+        ):
+            self.assertEquals(
+                test_xblock.has_support(getattr(test_xblock, view_name), functionality),
+                expected_result
+            )
```
doctoryes: What happens when `None` is the `functionality` checked in `has_support`?

xblock/test/test_runtime.py
```diff
@@ -524,10 +524,8 @@ def test_multiply_mixed(self):
         assert_equals(4, len(post_mixed.__bases__))
 
 
-@XBlock.needs("i18n")
-@XBlock.wants("secret_service")
-@XBlock.needs("no_such_service")
-@XBlock.wants("another_not_service")
+@XBlock.needs("i18n", "no_such_service")
+@XBlock.wants("secret_service", "another_not_service")
```
doctoryes: This is an improvement - but it did take a test away that checks that multiple `wants` or `needs` don't overwrite each other. From the implementation, it's clear that they don't. But just pointing it out.

xblock/mixins.py
```diff
@@ -538,3 +540,51 @@ def index_dictionary(self):
...
+            Internal decorator that updates the given view's list of supported
+            functionalities.
+            """
+            # pylint: disable=protected-access
+            if not hasattr(view, "_supports"):
+                view._supports = set()
+            for functionality in functionalities:
+                view._supports.add(functionality)
+            return view
+        return _decorator
```
doctoryes: There's nothing that enforces `functionalities` to be strings, which means integers, floats, lists, tuples, etc can all be passed in as functionalities. Is that the intent? Is it worth being concerned about?

nasthagiri: I'm not concerned about this, given we are using Python.  Do you really think we should be?

xblock/test/test_mixins.py
nasthagiri: It should return `False`.  Added a test for it.  Thanks.

xblock/mixins.py
doctoryes: Since you're not concerned, then I'm not concerned. :v Plus, you've added a comment for the intent.

doctoryes:
:+1:

PR #305, opened 2015-07-10
return empty string if value is none

Shrhawk:
[TNL-2686](https://openedx.atlassian.net/browse/TNL-2686)

If the xml attribute is setting using lxml library with ```None``` value then it raises ```TypeError``` ,  instead of passing ```None``` value to ```set(attribute, value)``` function , we can pass ```empty``` string (in case of NoneType) to avoid this error.

Test for the pr are at [#8889](https://github.com/edx/edx-platform/pull/8889)

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -780,10 +780,18 @@ class String(JSONField):
     MUTABLE = False
 
     def from_json(self, value):
-        if value is None or isinstance(value, basestring):
+        if value is None:
+            return ''
+        elif isinstance(value, basestring):
             return value
         else:
             raise TypeError('Value stored in a String must be None or a string, found %s' % type(value))
+
+    def to_json(self, value):
+        if value is None:
+            return ''
+        self._warn_deprecated_outside_JSONField()
+        return value
 
     def from_string(self, value):
         """String gets serialized and deserialized without quote marks."""
--- a/xblock/test/test_fields.py
+++ b/xblock/test/test_fields.py
@@ -212,7 +212,14 @@ class StringTest(FieldTest):
         self.assertJSONOrSetEquals('', '')
 
     def test_none(self):
-        self.assertJSONOrSetEquals(None, None)
+        # test for NoneType for to_string and from_string methods
+        test_field = String(enforce_type=True)
+
+        result_to_string = test_field.to_string(None)
+        self.assertEquals(result_to_string, '')
+
+        result_from_string = test_field.from_string(None)
+        self.assertEquals(result_from_string, '')
 
     def test_error(self):
         self.assertJSONOrSetTypeError(['a'])


Shrhawk:
@alawibaba @doctoryes @adampalay

xblock/mixins.py
```diff
@@ -522,7 +522,7 @@ def _add_field(self, node, field_name, field):
             elem.text = value
             node.insert(0, elem)
         else:
-            node.set(field_name, value)
+            node.set(field_name, '' if value is None else value)
```
adampalay: can you add a comment in the code why this is necessary and also add a test for this in this repo?

doctoryes: +1

Shrhawk:
test this please

xblock/mixins.py
```diff
@@ -522,7 +522,9 @@ def _add_field(self, node, field_name, field):
             elem.text = value
             node.insert(0, elem)
         else:
-            node.set(field_name, value)
+            # setting the None in the field_name raises TypeError to avoid this
```
adampalay: when does it raise a TypeError?

Shrhawk:
@adampalay  test and comment added

xblock/mixins.py
Shrhawk: when value of attribute is None e.g ```giturl``` which is ```setting``` field and if its not set will raise ```TypeError``` because of its ```None``` value

xblock/mixins.py
```diff
@@ -522,7 +522,9 @@ def _add_field(self, node, field_name, field):
             elem.text = value
             node.insert(0, elem)
         else:
-            node.set(field_name, value)
+            # setting the None in the value for node raises TypeError to avoid this
```
adampalay: why does it raise a `TypeError`?

xblock/test/test_parsing.py
```diff
@@ -450,6 +450,32 @@ def test_nested_roundtrip(self):
...
+        """
+        block = self.parse_xml_to_block("<leaf/>")
+
+        # example an setting field git_url  fo block
+        git_url = String(display_name="GIT URL", scope=Scope.settings, default=None)
+        git_url.__name__ = "git_url"
+
+        # set as mark as dirty so that we can set the value on node
+        # pylint: disable=protected-access
+        git_url._mark_dirty(block, git_url)
```
adampalay: rather than mark it as dirty, can't you just set it as another string?

Shrhawk: @adampalay  i can't understand the comment

xblock/mixins.py
Shrhawk: Its a expected behavior of lxml ```etree``` module

xblock/test/test_parsing.py
Shrhawk: @adampalay i tried but did'nt work

Shrhawk:
@adampalay @doctoryes

xblock/test/test_parsing.py
adampalay: @doctoryes , what's the right way to do this?

cpennington: It would probably be better to just make a new XBlock type that has the field you need, and then use that field for the test

xblock/mixins.py
cpennington: For OLX, if the value is the default for the field, then it shouldn't be set. Otherwise, it should be serialized to a string by the field. It looks like https://github.com/edx/XBlock/blob/master/xblock/mixins.py#L477-L478 takes care of not adding the field if it's a default. `field.to_string` should take care of the serialization. In what case is `field.to_string` returning `None`?

cpennington: Just to be clear, that seems like the bug to me. Why is `to_string` not returning a string?

xblock/test/test_parsing.py
```diff
@@ -450,6 +450,32 @@ def test_nested_roundtrip(self):
...
+
+        # add the setting field to git_url to block fields
+        block.fields['git_url'] = git_url
+
+        # pylint: disable=protected-access
+        block._field_data_cache['git_url'] = None
```
doctoryes: I don't understand what the previous three lines do. It seems like the code is unconventionally setting `git_url` to None. Why don't conventional means work? This code will add mystery - and no one wants to touch mysterious code after it's merged. The comments for this code should be more verbose.

xblock/mixins.py
Shrhawk: @cpennington 
 The ```to_string``` at [here](https://github.com/edx/XBlock/blob/017b46b80e712b1318379912257cf1cc1b68eb0e/xblock/fields.py#L792) call to ```to_json()``` [here](https://github.com/edx/XBlock/blob/017b46b80e712b1318379912257cf1cc1b68eb0e/xblock/fields.py#L563) which did'nt handle the ```None``` case so thats why its not returning string for ```None``` Type.

So should we overwrite  ```to_json``` for [class String](https://github.com/edx/XBlock/blob/017b46b80e712b1318379912257cf1cc1b68eb0e/xblock/fields.py#L773) with handling ```None``` case just like [here](https://github.com/edx/XBlock/blob/017b46b80e712b1318379912257cf1cc1b68eb0e/xblock/fields.py#L845) ? or should we handle ```None``` case at ```to_string``` method ?.

Shrhawk: @adampalay FYI

cpennington: Yes, I would override `to_json` and `from_json` in the `String` field to get sensible behavior that accounts for None.

Shrhawk:
@doctoryes @cpennington i have made changes kindly review .

xblock/fields.py
```diff
@@ -780,11 +780,19 @@ class String(JSONField):
...
+        elif isinstance(value, basestring):
             return value
         else:
             raise TypeError('Value stored in a String must be None or a string, found %s' % type(value))
 
+    def to_json(self, value):
+        if value is None:
+            return ''
+        self._warn_deprecated_outside_JSONField()
+        return value
```
cpennington: This is better, from the perspective of not causing bugs. It's a bit unfortunate that it effectively makes the semantics in OLX different than in code, since the serialized version can't actually support `None`. So, roundtripping `None` out to OLX and back turns it into `''`, unless `None` is the default value for the `String`.

cpennington: That said, I can't think of a better option, really. We're encoding `str|None` into `str`, so we're going to lose information somewhere, no matter what we do.

cpennington: So, reluctantly, :+1:.

adampalay:
why did coverage drop?

doctoryes:
:+1:

Shrhawk:
@adampalay i just squash the commits coverage is fine now

adampalay:
:+1:

sarina:
What's going on w/ this pr?

Shrhawk:
@sarina it ready to merge now

PR #306, opened 2015-07-13
WIP: shared field api review 07/13

beardeer:
@pmitros @cpennington @DongwookYoon

Hi team,

This is current "reviewable" shared field changes on the XBlock side. It works with current XBlock SDK, I haven't test it with devstack yet. 

As aways, please let me know your opinions. Thank you!

### DIFF
--- a/xblock/field_data.py
+++ b/xblock/field_data.py
@@ -11,7 +11,6 @@ from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 
 from xblock.exceptions import InvalidScopeError
-from xblock.core import XBlock
 
 class FieldData(object):
     """
--- a/xblock/field_data.py
+++ b/xblock/field_data.py
@@ -144,51 +144,17 @@ class SplitFieldData(FieldData):
         """
         self._scope_mappings = scope_mappings
 
-    def _find_scope_in_mapping(self, scope):
-        """Summary
-        
-        Args:
-            scope (TYPE): Description
-        
-        Returns:
-            TYPE: Description
-        """
-        if scope not in self._scope_mappings:
-            raise InvalidScopeError(scope)
-
-        return self._scope_mappings[scope]
-
     def _field_data(self, block, name):
         """Return the field data for the field `name` on the :class:`~xblock.core.XBlock` `block`"""
         scope = block.fields[name].scope
 
+        if scope not in self._scope_mappings:
+            raise InvalidScopeError(scope)
+
         return self._find_scope_in_mapping(scope)
 
-    def _shared_field_data(self, remote_scope):
-        """Summary
-        
-        Args:
-            remote_scope (TYPE): Description
-        
-        Returns:
-            TYPE: Description
-        
-        Deleted Parameters:
-            block (TYPE): Description
-            name (TYPE): Description
-        """
-        return self._find_scope_in_mapping(remote_scope)
-
     def get(self, block, name):
-        remote_scope = block.fields[name].query.remote_scope
-        if remote_scope is None:
-            return self._field_data(block, name).get(block, name)
-        else:
-            field_data_cache = self._shared_field_data(remote_scope)
-            try:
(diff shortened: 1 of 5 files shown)


xblock/fields.py
```diff
@@ -146,6 +148,24 @@ def scopes(cls):
...
 
 
+class SharedUserScope(object):
+    """
+    Enumeration of shared scopes.
+
+    To be continued.
+    """
+    ONE = Sentinel('SharedUserScope.ONE')
+    GROUP = Sentinel('SharedUserScope.GROUP')
```
pmitros: We don't have groups yet. I'd hold off on implementing this until we do, perhaps? I'm actually not entirely clear on why we need multiple scopes here in the first place -- shared implies it is not a user_state_summary, but per-user for now.

xblock/fields.py
```diff
@@ -233,6 +253,23 @@ def __eq__(self, other):
         return isinstance(other, Scope) and self.user == other.user and self.block == other.block
 
 
+class RemoteScope(Scope):
+    """
+    Defines types of remote scopes to be used.
+    """
+    shared_user_state = ScopeBase(SharedUserScope.ONE, BlockScope.USAGE, u'shared_user_state')
+    shared_group = ScopeBase(SharedUserScope.GROUP, BlockScope.USAGE, u'shared_group')
+    shared_platform = ScopeBase(SharedUserScope.ALL, BlockScope.USAGE, u'shared_platform')
```
pmitros: What is shared platform? It's not clear from the name. The way I see it, we want to probe across either user ids, usage ids or both?

xblock/query.py
```diff
@@ -0,0 +1,111 @@
...
+            xblock_class (TYPE): Description
+        
+        Returns:
+            TYPE: Description
+        """
+
+        return self._queryable
+
+    def __set__(self, field, val):
+        print val
```
pmitros: When there is dummy code like this, it is good to have a TODO comment, so you see a pylint warning/error if you try to commit it.

xblock/field_data.py
```diff
@@ -145,17 +144,51 @@ def __init__(self, scope_mappings):
...
+            TYPE: Description
+        
+        Deleted Parameters:
+            block (TYPE): Description
+            name (TYPE): Description
+        """
+        return self._find_scope_in_mapping(remote_scope)
+
     def get(self, block, name):
-        return self._field_data(block, name).get(block, name)
```
beardeer: I am re-considering the part of code changes,  I am trying to keep original XBlock code intact and push more changes to the platform side, but still trying to find out if that is feasible.

xblock/query.py
```diff
@@ -0,0 +1,111 @@
...
+        
+        Returns:
+            TYPE: Description
+        """
+        self._field_name = field_name
+        self._remote_scope = remote_scope
+        self._bind = bind
+
+    def _replace_xblock_user_id(self, xblock, user_id):
+        new_block = deepcopy(xblock)
```
pmitros: What is XBlock right here? Is it the actual XBlock object? If so, could a deepcopy like this may have unintended side effects? I'm not familiar enough with this code to know, but I'd like to flag for @cpennington to have a look.

xblock/runtime.py
```diff
@@ -237,8 +237,7 @@ def default(self, block, name):
         :param name: name of the field to default
         """
         return self._kvs.default(self._key(block, name))
-
-
+
```
pmitros: Please don't commit whitespace like this

pmitros:
I did a code review to the best of my ability. I'm not very familiar with the code around this. Looks sensible to me, but I'd rely more on @cpennington.

xblock/query.py
beardeer: What I need here a new scope id with querying user id and this new XBlock (deep-copied, so I don't change current runtime XBlock) is used to piggyback it to lower level calls.

PR #308, opened 2015-08-03
Make Field.to_string() and Field.from_string() methods more consistent.

smarnach:
**Description**

The old version of these methods was inherently assymmetric.  The deserialisation in `from_string()`
only called `from_json()` when `enforce_type()` was overridden to do so (which all classes with a
non-trivial `from_json()` implementation do) and `enable_enforce_type` was set to `True`.  This commit
makes the two functions more consistent by not considering `enable_enforce_type` at all for
deserialisation and calling `from_json()` in `enforce_type()` by default.  Serialisation and
deserialisation requires correct types to work already, so there is no harm in always enforciing
the type – the code fails in the same cases as before.

This commit fixes a bug in serialising `DateTime` fields.  The old version included spurious double
quotes around the string, which would not be correctly deserialised when `enable_enforce_type` was
not set (the default).

This patch was already [commited to the edx-solutions fork](https://github.com/edx-solutions/XBlock/pull/5) of this repostiory.
- - -
 
**JIRA Story** [OSPR-734](https://openedx.atlassian.net/browse/OSPR-734)
**Confluence / Product Asset** N/A
**Sandbox URL** [sandbox3.opencraft.com](http://sandbox3.opencraft.com/) – try importing some exported course tar.gz here, preferably one containing `DateTime` fields.
**Dependencies** N/A
**PR Author(s) Notes / To-Do** N/A
**Screenshots** N/A
- - -
**Reviewers**
- [x] Code: @e-kolpakov (thumbs up given in the [edx-solutions PR](https://github.com/edx-solutions/XBlock/pull/5))
- [ ] Code: (TBD)

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -439,7 +439,11 @@ class Field(Nameable):
                 value, traceback.format_exc().splitlines()[-1])
             warnings.warn(message, FailingEnforceTypeWarning, stacklevel=3)
         else:
-            if value != new_value:
+            try:
+                equal = value == new_value
+            except TypeError:
+                equal = False
+            if not equal:
                 message = u"The value {} would be enforced to {}".format(
                     value, new_value)
                 warnings.warn(message, ModifyingEnforceTypeWarning, stacklevel=3)
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -602,7 +602,7 @@ class Field(Nameable):
         """
         self._warn_deprecated_outside_JSONField()
         value = yaml.safe_load(serialized)
-        return self._check_or_enforce_type(value)
+        return self.enforce_type(value)
 
     def enforce_type(self, value):
         """
@@ -614,7 +614,7 @@ class Field(Nameable):
         This must not have side effects, since it will be executed to trigger
         a DeprecationWarning even if enforce_type is disabled
         """
-        return value
+        return self.from_json(value)
 
     def read_from(self, xblock):
         """
@@ -678,8 +678,6 @@ class Integer(JSONField):
             return None
         return int(value)
 
-    enforce_type = from_json
-
 
 class Float(JSONField):
     """
@@ -696,8 +694,6 @@ class Float(JSONField):
         if value is None or value == '':
             return None
         return float(value)
-
-    enforce_type = from_json
 
 
 class Boolean(JSONField):
@@ -735,8 +731,6 @@ class Boolean(JSONField):
         else:
             return bool(value)
 
-    enforce_type = from_json
-
 
 class Dict(JSONField):
     """
@@ -753,8 +747,6 @@ class Dict(JSONField):
         else:
             raise TypeError('Value stored in a Dict must be None or a dict, found %s' % type(value))
 
-    enforce_type = from_json
-
 
 class List(JSONField):
     """
@@ -770,8 +762,6 @@ class List(JSONField):
             return value
         else:
             raise TypeError('Value stored in a List must be None or a list, found %s' % type(value))
-
-    enforce_type = from_json
 
 
 class Set(JSONField):
@@ -825,8 +815,6 @@ class String(JSONField):
         """String gets serialized and deserialized without quote marks."""
         return self.to_json(value)
 
-    enforce_type = from_json
-
 
 class DateTime(JSONField):
     """
@@ -842,31 +830,27 @@ class DateTime(JSONField):
         """
         Parse the date from an ISO-formatted date string, or None.
         """
-        if isinstance(value, basestring):
-
-            # Parser interprets empty string as now by default
-            if value == "":
-                return None
-
-            try:
-                parsed_date = dateutil.parser.parse(value)
-            except (TypeError, ValueError):
-                raise ValueError("Could not parse {} as a date".format(value))
-
-            if parsed_date.tzinfo is not None:  # pylint: disable=maybe-no-member
-                parsed_date.astimezone(pytz.utc)  # pylint: disable=maybe-no-member
-            else:
-                parsed_date = parsed_date.replace(tzinfo=pytz.utc)  # pylint: disable=maybe-no-member
-
-            return parsed_date
-
         if value is None:
             return None
 
-        if isinstance(value, datetime.datetime):
-            return value
+        if isinstance(value, basestring):
+            # Parser interprets empty string as now by default
+            if value == "":
+                return None
+            try:
+                value = dateutil.parser.parse(value)
+            except (TypeError, ValueError):
+                raise ValueError("Could not parse {} as a date".format(value))
 
-        raise TypeError("Value should be loaded from a string, not {}".format(type(value)))
+        if not isinstance(value, datetime.datetime):
+            raise TypeError(
+                "Value should be loaded from a string, a datetime object or None, not {}".format(type(value))
+            )
+
+        if value.tzinfo is not None:  # pylint: disable=maybe-no-member
+            return value.astimezone(pytz.utc)  # pylint: disable=maybe-no-member
+        else:
+            return value.replace(tzinfo=pytz.utc)  # pylint: disable=maybe-no-member
 
     def to_json(self, value):
         """
@@ -878,11 +862,9 @@ class DateTime(JSONField):
             return None
         raise TypeError("Value stored must be a datetime object, not {}".format(type(value)))
 
-    def enforce_type(self, value):
-        if isinstance(value, datetime.datetime) or value is None:
-            return value
-
-        return self.from_json(value)
+    def to_string(self, value):
+        """DateTime fields get serialized without quote marks."""
+        return self.to_json(value)
 
 
 class Any(JSONField):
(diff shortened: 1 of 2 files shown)


edx-webhook:
Thanks for the pull request, @smarnach! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2FXBlock&number=308

edx-webhook:
Thanks for the pull request, @smarnach! I've created [OSPR-734](https://openedx.atlassian.net/browse/OSPR-734) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

We can't start reviewing your pull request until you've added yourself to the [AUTHORS](https://github.com/edx/XBlock/blob/master/AUTHORS) file. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

tusbar:
@smarnach @Kelketek you may want to rebase and include changes to the Set field that was included in PR https://github.com/edx/XBlock/pull/311

Kelketek:
@tusbar I remember reading something about #311 's contents being accidentally removed via force push. Are they supposed to be in Cypress? Need to know for purposes elsewhere.

smarnach:
@tusbar Thanks – I rebased on top of current master.

@sarina Could you please take a look at this PR, please?  (Not sure if I should ping you on OSPRs, or whether just creating the JIRA ticket is enough.  This isn't urgent.)

sarina:
@smarnach unfortunately I don't have the expertise to review this. Creating the JIRA ticket is enough for the review and if you want to understand the status of the review you should go visit the JIRA ticket, and perhaps bug people on that ticket.

Kelketek:
@smarnach While my patch fixes some cases where this throws the LMS a curveball, it doesn't fix all of them. I've run out of steam for the day, but since this change is not going to be mergeable for upstream until the issues are fixed (as I believe the same failures would happen there as well), and it's blocking the rebase, please try to prioritize fixing this as soon as you start work.

The test failures are in `xmodule/modulestore/tests/test_split_modulestore.py`

And look like:

```
42) ERROR: test_schema (xmodule.modulestore.tests.test_split_modulestore.TestSchema)

    common/lib/xmodule/xmodule/modulestore/tests/test_split_modulestore.py line 2060 in test_schema
      db_connection = modulestore().db_connection
    common/lib/xmodule/xmodule/modulestore/tests/test_split_modulestore.py line 2101 in modulestore
      SplitModuleTest.bootstrapDB(SplitModuleTest.modulestore)
    common/lib/xmodule/xmodule/modulestore/tests/test_split_modulestore.py line 525 in bootstrapDB
      course = split_store.persist_xblock_dag(course, revision['user_id'])
    common/lib/xmodule/xmodule/modulestore/split_mongo/split.py line 2005 in persist_xblock_dag
      return self.get_item(xblock.location.for_version(new_id))
    /usr/lib/python2.7/contextlib.py line 24 in __exit__
      self.gen.next()
    common/lib/xmodule/xmodule/modulestore/__init__.py line 185 in bulk_operations
      self._end_bulk_operation(course_id, emit_signals)
    common/lib/xmodule/xmodule/modulestore/__init__.py line 273 in _end_bulk_operation
      dirty = self._end_outermost_bulk_operation(bulk_ops_record, structure_key)
    common/lib/xmodule/xmodule/modulestore/split_mongo/split.py line 246 in _end_outermost_bulk_operation
      self.db_connection.insert_structure(bulk_write_record.structures[_id], bulk_write_record.course_key)
    common/lib/xmodule/xmodule/modulestore/split_mongo/mongo_connection.py line 407 in insert_structure
      self.structures.insert(structure_to_mongo(structure, course_context))
    /edx/app/edxapp/venvs/edxapp/src/mongodb-proxy/mongodb_proxy.py line 115 in __call__
      return self.proxied_object(*args, **kwargs)
    /edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/pymongo/collection.py line 409 in insert
      gen(), check_keys, self.uuid_subtype, client)
   InvalidDocument: Cannot encode object: <function link_func at 0xd431c08>
```

XBlock seems to be attempting to serialize Course Tabs when it normally would not have.

@antoviaque FYI

xblock/fields.py
```diff
@@ -838,31 +842,27 @@ def from_json(self, value):
...
-        if isinstance(value, datetime.datetime):
-            return value
+        if not isinstance(value, datetime.datetime):
+            raise TypeError(
+                "Value should be loaded from a string, a datetime object or None, not {}".format(type(value))
+            )
 
-        raise TypeError("Value should be loaded from a string, not {}".format(type(value)))
+        if value.tzinfo is not None:  # pylint: disable=maybe-no-member
+            return value.astimezone(pytz.utc)  # pylint: disable=maybe-no-member
```
doctoryes: Before this change, it looks like `.astimezone()` was actually doing nothing, right? Or did `parsed_date` actually get its time zone changed? The way you've done it is definitely more correct - I just wonder about side-effects of code relying on the previous behavior. Perhaps you should add a test which sets a DateTime field to a value with a timezone different than UTC - does one exist now?

sarina: How does this interact with edx platform, where afaik there is no support
yet for timezones and everything is represented in the server's time?

xblock/test/test_fields.py
```diff
@@ -780,7 +810,11 @@ def test_to_string_regexp_matches(self, _type, value, regexp):
                   kaw: null
             """),
             [1, 2.345, {"foo": True, "bar": [1, 2, 3]}, {"meow": False, "woof": True, "kaw": None}]
-        )
+        ),
+        # Test that legacy DateTime format including double quotes can still be imported for compatibility with
+        # old data export tar balls.
+        (DateTime, '"2014-04-01T02:03:04.567890"', dt.datetime(2014, 4, 1, 2, 3, 4, 567890).replace(tzinfo=pytz.utc)),
+        (DateTime, '"2014-04-01T02:03:04.000000"', dt.datetime(2014, 4, 1, 2, 3, 4).replace(tzinfo=pytz.utc)),
     )
```
doctoryes: Since a datetime string *without* double quotes should also pass this test, perhaps add a couple of those to underscore that fact? I know that it's tested in `test_both_directions` above. I'm not even sure why `test_from_string` can't just be combined into `test_both_directions`. This new test case is perhaps the only one that can't be combined into `test_both_directions` as it won't be symmetric.

smarnach: @doctoryes These two functions essentially test the same thing, yes.  The tests are structured in two groups since some of them work both ways, whereas others only test deserialisation of strings that we should be able to deserialise, but that would not be serialised in exactly the way specified in the test data.  All test data for `test_both_directions()` falls into the first category, and all test data for `test_from_string()` falls into the second category, for various reasons.  Some of them test YAML code that isn't JSON, others test non-standard capitalisation or number formats.

I added the `DateTime` tests to the two functions according to the same principles:  The ones that are symmetric go in the first function, the ones that only test correct deserialisation go in the second function.  Maybe I could add a comment to make the idea more obvious?

xblock/fields.py
smarnach: The main problem with the old implementation was that it correctly adjusted the timezone of values passed in as string, but did not adjust the timezone of values passed in as `datetime` instances.  This results in rather inconsistent behaviour, since `yaml.safe_load()`, which is used as first step when deserialising data, already parses unquoted date strings into `datetime` instances, whereas quoted date string are returned as strings, which will only be parsed into `datetime` instances in this function.  In essence, adding and removing quotes resulted in different handling of timezones.  I don't think there is any code out there relying on this bug, but I've been wrong with this kind of assumption before (see https://xkcd.com/1172/).  For what it's worth, all edx-platform tests pass with this patch applied.

@doctoryes I think `.astimezone()` did change the timezone in some cases even before this change, but it wasn't applied when the value passed in already was a `datetime` instance.

@sarina Not sure about this.  This fix doesn't change the treatment of timezones fundamentally; it only makes it behave consistently.  Maybe edx-platform only works correctly if the server time is UTC?  I really don't know.

xblock/test/test_fields.py
doctoryes: Yes - a comment above the `test_from_string` ddt data would be nice - basically saying what you've said above about the data.

xblock/fields.py
doctoryes: I also ran these XBlock changes through the edx-platform test suite today - and they passed.

doctoryes:
After the comment is added, :+1:  I'll be out the next couple of days - but I'll inform a team member to merge the PR after its addition.

smarnach:
@doctoryes Thanks – I'll add the comment tomorrow, since I'm logging off for today now.  And don't worry about merging – I can handle that myself (including updating the commit hash on edx-platform).

doctoryes:
@smarnach OK - thanks!

Issue #309, opened 2015-08-05
What happened to the Set field?


tusbar:
Hey,

A while back I added support for `Set` type fields. The PR was merged, but it seems to have been removed altogether as the commits are not even here anymore.

Here’s the PR: https://github.com/edx/XBlock/pull/212

Was something wrong with it so it had to be removed?

I had a XBlock that was depending on it.

cc @dmitchell @cpennington @sarina

sarina:
I'm so confused. I don't see the merged commits in the repo history either: https://github.com/edx/XBlock/commits/master?page=4

tusbar:
@sarina it looks like the repo has been force-pushed and the commit (possibly others?) was left out.

sarina:
oh what?? that's not good - who knows what other commits may have been missed :disappointed: 

Can you make a new pull request with your changes? I'm sure we can get them merged promptly.

sarina:
@tusbar  i don't suppose there are forensics to see a force push?  It'd be nice if github noted it in the history

tusbar:
@sarina  The objects are here, just not included in master: https://github.com/edx/XBlock/commit/8124106ffeced4d12f303e6eac9d8dfa34d27dd4

I’ll do another PR.

sarina:
Cool. We're doing some git magic over here to try to figure out if other commits are missing as well. ugh.

sarina:
@tusbar FYI I've made the PR ^^ we couldn't find anything else that is missing, and unfortunately we can't figure out who force pushed. However we've disabled force pushing on master for all edx/ repos going forward.

tusbar:
@sarina haha, I’ve done it too.

tusbar:
Tests will pass on mine though. :wink:

tusbar:
@sarina how did you disable force pushing? I’ve been looking for that setting on GitHub for years.

sarina:
@tusbar our ops team got in touch with Github directly.

We reached out to github, and they've helpfully put a server side pre-receive hook that blocks force pushes to the default branch (typically 'master') of all of our repos.  Attempting to do a git push --force against master will now result in the following error message:

`! [remote rejected] HEAD -> master (pre-receive hook declined)`

tusbar:
@sarina cool, I’ll try and do that as well. :)

smarnach:
[...]

PR #311, opened 2015-08-05
Add Set field

tusbar:
See #309.

ping @sarina @cpennington


sarina:
I think it makes sense to just merge this, right? cc @nedbat

tusbar:
:+1:

sarina:
Since this was already reviewed in #212 and is passing tests, @tusbar I think go ahead and merge it!

PR #312, opened 2015-08-06
(WIP) Make Field.to_string() and Field.from_string() methods more consistent.

Kelketek:
The old version of these methods was inherently assymmetric.  The deserialisation in from_string()
only called from_json() when enforce_type() was overridden to do so (which all classes with a
non-trivial from_json() implementation do) and enable_enforce_type was set to True.  This commit
makes the two functions more consistent by not considering enable_enforce_type at all for
deserialisation and calling from_json() in enforce_type() by default.  Serialisation and
deserialisation requires correct types to work already, so there is no harm in always enforciing
the type – the code fails in the same cases as before.

This commit fixes a bug in serialising DateTime fields.  The old version included spurious double
quotes around the string, which would not be correctly deserialised when enable_enforce_type was
not set (the default).

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -599,7 +599,7 @@ class Field(Nameable):
         """
         self._warn_deprecated_outside_JSONField()
         value = yaml.safe_load(serialized)
-        return self._check_or_enforce_type(value)
+        return self.enforce_type(value)
 
     def enforce_type(self, value):
         """
@@ -611,7 +611,7 @@ class Field(Nameable):
         This must not have side effects, since it will be executed to trigger
         a DeprecationWarning even if enforce_type is disabled
         """
-        return value
+        return self.from_json(value)
 
     def read_from(self, xblock):
         """
@@ -695,8 +695,6 @@ class Integer(JSONField):
             return None
         return int(value)
 
-    enforce_type = from_json
-
 
 class Float(JSONField):
     """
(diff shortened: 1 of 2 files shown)


edx-webhook:
Thanks for the pull request, @Kelketek! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2FXBlock&number=312

smarnach:
Closing this in favour of the identical #308.

Kelketek:
Re-opening to exploit CI build because I'm having trouble getting that to work locally. Will re-close.

smarnach:
:+1:

Kelketek:
@smarnach Made another change to this, which you will probably want to cherry-pick into your PR. It is accompanied by the change made in the solutions platform fork, `5271e287fd3ee97d3590a6dc58c81600a7336928`, which fixes a bug your PR exposes.

bradenmacdonald:
:+1: I can't test this completely due to some discussion related errors in the LMS, but this looks good and seems to be working.

smarnach:
:+1: For the additional commit.

PR #315, opened 2015-08-07
Add pypi version badge

tusbar:
Just because it’s cool.

cc @sarina, @singingwolfboy

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,8 +2,8 @@ Part of `edX code`__.
 
 __ http://code.edx.org/
 
-XBlock Courseware Components |build-status| |coverage-status|
-=============================================================
+XBlock Courseware Components |build-status| |coverage-status| |pypi-version|
+============================================================================
 
 XBlock is a component architecture by edX.org for building courseware.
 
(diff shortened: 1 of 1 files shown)


singingwolfboy:
Sounds good to me. :+1:

sarina:
What do we have on PyPi? @singingwolfboy do you know who is responsible for pushing pypi versions of this repo? AFAIK we're not incorporating pypi into our dev cycles (we _should_ but so far we havne't)

tusbar:
The issue with this is that the badges won’t all stay on the same line. :(

singingwolfboy:
@sarina I don't think anyone at edX is responsible for pushing PyPI versions of the repo right now. This seems like something that would fall to the platform team.

sarina:
@tusbar I'm not sure we want to advertise this badge until we're actively using pypi to distribute the package.

tusbar:
@sarina alright! I thought it was up to date!

sarina:
It's last updated November 2014. https://pypi.python.org/pypi/XBlock/0.4

tusbar:
Yes indeed. It would be nice to have dependencies like XBlock on pypi. So `pip install` would not spend so much unnecessary time resolving github refs.

sarina:
I agree! I'll put a ticket to start doing this for xblock.

PR #316, opened 2015-08-12
Revert "String field return empty string for None"

Shrhawk:
[PLAT-736](https://openedx.atlassian.net/browse/PLAT-736)

Reverted the related pr [#8889](https://github.com/edx/edx-platform/pull/8889) on edx -platform produce issue [TNL-3031](https://openedx.atlassian.net/browse/TNL-3031).

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -808,18 +808,10 @@ class String(JSONField):
     MUTABLE = False
 
     def from_json(self, value):
-        if value is None:
-            return ''
-        elif isinstance(value, basestring):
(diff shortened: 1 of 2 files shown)


Shrhawk:
@adampalay @waheedahmed this will revert https://github.com/edx/XBlock/pull/305

adampalay:
@cpennington 

:+1:

cpennington:
:+1:

cpennington:
It would be nice to move this fix to `to_string` and `from_string` to fix the original issue with XML import/export.

PR #317, opened 2015-08-13
WIP: shared fields - XBlock review

beardeer:
Hi team @cpennington and @pmitros,

Here is my changes on the XBlock side. A summery of changes in each file: 

exceptions.py: Adding a new exception to rise access deny error for shared fields. 
fields.py: Attaching Query and Shared objects to Field
query.py: The implementation details of Queryable object (used by Query and Shared)
runtime.py: Adding a new method for creating a XBlock by user id and usage id

Also, the documentation that talks about the high level design can be found out here. 
https://docs.google.com/document/d/1l3vV-gwbipJC5OA_aC4kyAdh84JxEjiz5Ni_X8iOhuw/edit#


xblock/query.py
```diff
@@ -0,0 +1,226 @@
...
+class Query(object):
+    """ 
+    A query object for accessing shared field in XBlock with Queryable operators
+
+    Parameters:
+
+        field_name (str): the name of the shared field
+    """
+    def __init__(self, field_name):
+        self._queryable = Queryable(field_name=field_name)
```
cpennington: `Query` is going to be instantiated on a per-xblock-class basis, right? Which means that `_queryable` will be as well. I worry about having the "current" xblock attached to a class-global object.

Could you instead instantiate a new `Queryable` in `__get__`?

xblock/fields.py
```diff
@@ -646,6 +650,24 @@ def is_set_on(self, xblock):
 
     def __hash__(self):
         return hash(self.name)
+    
+    @classmethod
+    def Query(cls, field_name):
+        """
+        Return a instance of Query which allow ad-hoc queries to access shared fields. 
+        """
+        return Query(field_name)
```
cpennington: Doesn't `Query` need to get passed `cls`, so that it knows what type of field it's being declared as? (Although, thinking about it now, when does it resolve issues where the type of `cls` and the type of the field identified by `field_name` differ?)

Also, `field_name` isn't enough to uniquely identify a block, is it? How do you reference a field in another XBlock type? Does the `Query` definition need to have the block type and the field name passed as arguments?

beardeer: field_name is not enough to find a block, so the get and set methods of Query need both user id and usage id in order to local a block.

xblock/query.py
beardeer: Workin on it.

xblock/fields.py
beardeer: Thoughts on type of cls and type of querying field: if we instantiate Query as an independent object instead of attaching it to Field, we can stop worry about mismatch types and Query uses whatever the querying field defines.

cpennington: re `field_name`: Pushing it further, couldn't you just pass the field_name at the time that you execute the query? In which case, maybe `Query` should just be an XBlock service, and `Shared` is the only field-like thing.

Alternately, you could allow the user to pass a whole set of things to the `Query` object, to pre-seed (and allow the app to pre-load) as much as possible. So, if you gave it a block_type and a field_name, then when you query, it can check that the usage you passed is the block type you were expecting (for instance).

beardeer: Yes, I agree that Query can just be an XBlock service, instead of an inner
class of Field. I will see how far I can proceed on this path.

On Thu, Aug 13, 2015 at 4:15 PM, Calen Pennington <notifications@github.com>
wrote:

> In xblock/fields.py
> <https://github.com/edx/XBlock/pull/317#discussion_r37019946>:
>
> > @@ -646,6 +650,24 @@ def is_set_on(self, xblock):
> >
> >      def __hash__(self):
> >          return hash(self.name)
> > +
> > +    @classmethod
> > +    def Query(cls, field_name):
> > +        """
> > +        Return a instance of Query which allow ad-hoc queries to access shared fields.
> > +        """
> > +        return Query(field_name)
>
> re field_name: Pushing it further, couldn't you just pass the field_name
> at the time that you execute the query? In which case, maybe Query should
> just be an XBlock service, and Shared is the only field-like thing.
>
> Alternately, you could allow the user to pass a whole set of things to the
> Query object, to pre-seed (and allow the app to pre-load) as much as
> possible. So, if you gave it a block_type and a field_name, then when you
> query, it can check that the usage you passed is the block type you were
> expecting (for instance).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/edx/XBlock/pull/317/files#r37019946>.
>



-- 
Sincerely,
Xiaolu Xiong
http://www.wpi.edu/~xxiong/

beardeer:
We are not actively working on this one, I am closing this PR for now.

pmitros:
@cpennington You were considering bringing this one over the finish line. Any chance?

cpennington:
Yes, I'm planning to. Just haven't gotten back to it.

PR #320, opened 2015-09-17
Upgrade django version to 1.8.

muzaffaryousaf:
TNL-3299

### DIFF
--- a/django_requirements.txt
+++ b/django_requirements.txt
@@ -1,3 +1,3 @@
 # Install Django requirements, if we're using the optional Django-integrated
 # parts of XBlock
-Django >= 1.4, < 1.7
+Django >= 1.8, < 1.9


muzaffaryousaf:
@nedbat Can you please review this ? Any suggestions/testing strategies are most welcome. 
Thanks

muzaffaryousaf:
@symbolist @macdiesel Please give it a look. Thanks.

django_requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Install Django requirements, if we're using the optional Django-integrated
 # parts of XBlock
-Django >= 1.4, < 1.7
\ No newline at end of file
+Django >= 1.8, < 1.9
```
nedbat: Since this is a library intended for wide use, I would keep the requirement as broad as possible.  We can leave it as `>= 1.4, < 1.9`, can't we?

nedbat:
For testing: there are django tests in the XBlock repo.  I guess we don't have good automation here to make sure they run.  But if you install django-requirements.txt, then the django tests will run (and pass).

symbolist:
@nedbat In the Travis logs, I can see django-requirements.txt being installed so the django tests should be running too.

Nothing more from my side. :+1:

django_requirements.txt
macdiesel: +1 on this

PR #321, opened 2015-09-18
Add labels to each topic

mhoeber:
@cpennington  or @nedbat   small addition that isn't user visible.  We are doing this across all doc to be able to use intersphinx to manage an obscene amount of links without hardcoding URLs.


cpennington:
:+1:

PR #322, opened 2015-10-14
Hackathon: AMD support for XBlocks

andy-armstrong:


andy-armstrong:
@e-kolpakov @cahrens

xblock/fragment.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, content=None):
         self.js_init_fn = None
         self.js_init_version = None
         self.json_init_args = None
+        self.use_require_js = False
```
andy-armstrong: I wonder if we should call this ``use_amd`` so that other runtimes can use an alternative to RequireJS if they choose.

xblock/fragment.py
```diff
@@ -181,7 +182,7 @@ def add_frags_resources(self, frags):
         for resource in frags:
             self.add_frag_resources(resource)
 
-    def initialize_js(self, js_func, json_args=None):
+    def initialize_js(self, js_func, json_args=None, use_require_js=False):
```
andy-armstrong: I'm still conflicted as to whether this would read more cleanly for the client if it was called ``initialize_amd``. If not, then the first parameter should be renamed from ``js_func`` to make it clearer that it can be either a function or a RequireJS definition that returns a function.

xblock/fragment.py
```diff
@@ -181,7 +182,7 @@ def add_frags_resources(self, frags):
...
         """Register a Javascript function to initialize the Javascript resources.
 
         `js_func` is the name of a Javascript function defined by one of the
```
andy-armstrong: This docstring should be updated to describe the new RequireJS behavior.

xblock/runtime.py
```diff
@@ -886,6 +886,7 @@ def _wrap_ele(self, block, view, frag, extra_data=None):
         if frag.js_init_fn:
             data['init'] = frag.js_init_fn
             data['runtime-version'] = frag.js_init_version
+            data['use-require'] = frag.use_require_js
```
andy-armstrong: I think this should probably be ``use-amd``.

andy-armstrong:
As we discussed, we need some tests to ensure that we have good coverage of the new logic.

xblock/runtime.py
```diff
@@ -886,6 +886,7 @@ def _wrap_ele(self, block, view, frag, extra_data=None):
         if frag.js_init_fn:
             data['init'] = frag.js_init_fn
```
andy-armstrong: We need to think how to pass down the static file when this is being run in DEBUG mode, so that it can be served from the CDN. This will only be possible once this PR lands: https://github.com/edx/edx-platform/pull/10176

Another thing to note: XBlocks will serve uncached files in production with this change until we can work out how to use Optimizer. This is because RequireJS is unaware of the MD5 hashes and so will load files without them. These files are deliberately not cached in production because there is no way to know whether the content has changed. This is why we introduced Optimizer into the LMS in the first place -- see https://openedx.atlassian.net/wiki/display/TNL/RequireJS+in+the+LMS#RequireJSintheLMS-Background

edx-webhook:
Thanks for the pull request, @andy-armstrong! I've created [OSPR-2114](https://openedx.atlassian.net/browse/OSPR-2114) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/XBlock/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

PR #323, opened 2015-10-20
Adding message in XBlockSaveError Exception

awaisdar001:
[TNL-2324](https://openedx.atlassian.net/browse/TNL-2324)

Adding message in `XBlockSaveError` excception. Current exception in [mixins.py#L275](https://github.com/edx/XBlock/blob/111935cab69a9128948180e6b9673c391d2ff165/xblock/mixins.py#L275) doesn't log the fields on which the error is occurring. 

#### Example of current logging of exception
```
File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/newrelic-2.46.0.37/newrelic/hooks/framework_django.py", line 499, in wrapper return wrapped(*args, **kwargs)
File "/edx/app/edxapp/edx-platform/lms/djangoapps/courseware/module_render.py", line 945, in handle_xblock_callback return _invoke_xblock_handler(request, course_id, usage_id, handler, suffix, course=course)
File "/edx/app/edxapp/edx-platform/lms/djangoapps/courseware/module_render.py", line 1064, in _invoke_xblock_handler resp = instance.handle(handler, req, suffix)
File "/edx/app/edxapp/venvs/edxapp/src/xblock/xblock/mixins.py", line 86, in handle return self.runtime.handle(self, handler_name, request, suffix)
File "/edx/app/edxapp/edx-platform/common/lib/xmodule/xmodule/x_module.py", line 1274, in handle return super(MetricsMixin, self).handle(block, handler_name, request, suffix=suffix)
File "/edx/app/edxapp/venvs/edxapp/src/xblock/xblock/runtime.py", line 1035, in handle block.save()
File "/edx/app/edxapp/venvs/edxapp/src/xblock/xblock/mixins.py", line 253, in save self.force_save_fields(fields_to_save)
File "/edx/app/edxapp/venvs/edxapp/src/xblock/xblock/mixins.py", line 275, in force_save_fields raise XBlockSaveError(saved_fields, fields)
XBlockSaveError
```

When there will be an error (e.g. `DatabaseError` in [lms/djangoapps/courseware/model_data.py#L421-L428](https://github.com/edx/edx-platform/blob/062e979e0ed14161d0f882091a06e032ca76bb3e/lms/djangoapps/courseware/model_data.py#L421-L428) and `KeyValueMultiSaveError`  exceptions is raised, we can excact the following logging. 

Example of logging after change
-
#### Code
```
raise KeyValueMultiSaveError(['error_field'])
```
####  Log
```
    self.force_save_fields(fields_to_save)
  File "/edx/app/edxapp/venvs/edxapp/src/xblock/xblock/mixins.py", line 277, in force_save_fields
    raise XBlockSaveError(saved_fields, fields, msg)
XBlockSaveError: Error saving fields ['error_field']
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 # 3rd-party needs
+pyyaml
 lxml
 requests
 webob
--- a/xblock/exceptions.py
+++ b/xblock/exceptions.py
@@ -21,7 +21,7 @@ class XBlockSaveError(Exception):
     """
     Raised to indicate an error in saving an XBlock
     """
-    def __init__(self, saved_fields, dirty_fields):
+    def __init__(self, saved_fields, dirty_fields, message=None):
         """
         Create a new XBlockSaveError
 
@@ -30,7 +30,7 @@ class XBlockSaveError(Exception):
         `dirty_fields` - a set of fields that were left dirty after the save
         """
         # Exception is an old-style class, so can't use super
-        Exception.__init__(self)
+        Exception.__init__(self, message)
 
         self.saved_fields = saved_fields
         self.dirty_fields = dirty_fields
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -272,7 +272,8 @@ class ScopedStorageMixin(RuntimeServicesMixin):
                 fields.remove(field)
                 # if the field was dirty, delete from dirty fields
                 self._reset_dirty_field(field)
-            raise XBlockSaveError(saved_fields, fields)
+            msg = 'Error saving fields {}'.format(save_error.saved_field_names)
+            raise XBlockSaveError(saved_fields, fields, msg)
 
         # Remove all dirty fields, since the save was successful
         for field in fields:
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,3 +35,4 @@ David Bodor <david.gabor.bodor@gmail.com>
 Dhruv Baldawa <dhruvbaldawa@gmail.com>
 Matjaz Gregoric <mtyaka@gmail.com>
 Adam Palay <adam@edx.org>
+Awais Jibran <awaisdar001@gmail.com>


awaisdar001:
@doctoryes @adampalay Please review.
Adding some logging; please suggest how else it can be improved.

xblock/exceptions.py
```diff
@@ -21,7 +21,7 @@ class XBlockSaveError(Exception):
     """
     Raised to indicate an error in saving an XBlock
     """
-    def __init__(self, saved_fields, dirty_fields):
+    def __init__(self, saved_fields, dirty_fields, message=None, *args):
```
adampalay: why do you have the `*args` variable here? I don't see why it's necessary.

xblock/exceptions.py
```diff
@@ -30,8 +30,9 @@ def __init__(self, saved_fields, dirty_fields):
         `dirty_fields` - a set of fields that were left dirty after the save
         """
         # Exception is an old-style class, so can't use super
-        Exception.__init__(self)
+        Exception.__init__(self, message, *args)
```
adampalay: why are we putting `message` in this `__init__`?

xblock/mixins.py
```diff
@@ -272,7 +272,8 @@ def force_save_fields(self, field_names):
                 fields.remove(field)
                 # if the field was dirty, delete from dirty fields
                 self._reset_dirty_field(field)
-            raise XBlockSaveError(saved_fields, fields)
+            msg = 'Error saving fields {}'.format([key for key in fields])
```
adampalay: do we want to log the values as well?

xblock/exceptions.py
awaisdar001: I used `*args` if user wanted to pass any other arguments. basically allow users initialize misc.
When i saw the signature of `Exception` (super class of XBlockSaveError) its `__init__` signature was accepting `*args` and `**kwargs`
like this

`def __init__(self, *args, **kwargs)`

We can either add or remove both; which makes more sense. Thoughts?

xblock/mixins.py
awaisdar001: Yes; `fields` are the list of `dirty_fields`, so we can log to get a sense which fields are causing the problem.

adampalay: Oh, sorry, I thought `fields` was a dictionary

xblock/exceptions.py
adampalay: let's leave out both, and if people want to extend it further, we can add that in a future PR

awaisdar001:
@adampalay  I have updated the code, and description; how the exception would look like.

xblock/exceptions.py
```diff
@@ -30,8 +30,9 @@ def __init__(self, saved_fields, dirty_fields):
         `dirty_fields` - a set of fields that were left dirty after the save
         """
         # Exception is an old-style class, so can't use super
-        Exception.__init__(self)
+        Exception.__init__(self, message)
```
adampalay: sorry, still unclear why we're passing `message` here

awaisdar001: Oh, I thought i added the comment earlier, We are setting the message attribute to the `Exception`
E.g.
```
try:
    1/0
except Exception as e:
    log(e.message)
```
So we have to set the message class attribute.

awaisdar001: Still unclear, please see this.
[xblock/exceptions.py](https://github.com/edx/XBlock/blob/111935cab69a9128948180e6b9673c391d2ff165/xblock/exceptions.py#L17)

adampalay: so then why set it again at L35?

awaisdar001: Pardon, Now I understand. Either of them is enough. So let go with keeping it in the `__init__`?
What do you think?

adampalay: aesthetically, I like setting it outside the `__init__`, but up to you :)

awaisdar001:
Updated, Now it will log with the fields. 

Note: The log depends on  `KeyValueMultiSaveError(param)` exception. 

Please review.

adampalay:
:+1: 
@doctoryes , may you please review too?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 # 3rd-party needs
+pyyaml
```
stvstnfrd: Why the added dependency? Is this a debugging relic?

adampalay: @stvstnfrd tests were failing without it

adampalay: Oh, but `yml` doesn't appear in the code at all :/

doctoryes: Looks like fields.py imports `yaml` - that's why.

doctoryes:
:+1:

PR #326, opened 2015-11-20
Force export runtime option + some tests for XmlSerializationMixin

e-kolpakov:
**Description:** This PR adds an option to force export any field. Forward-port of https://github.com/edx-solutions/XBlock/pull/6

The motivation is the following: `UNIQUE_ID` sentinel is used to generate a unique pseudo-random value for an XBlock field. The value generated is not actually random, and is stable (i.e. subsequent calculations return same value for given field and XBlock instances). The problem though, is that this value is not persisted in export-import workflow.

Direct motivation, more details and discussion of various approaches can be seen in the appropriate [JIRA task](https://edx-wiki.atlassian.net/browse/MCKIN-3637)

**JIRA Task:** [MCKIN-3637](https://edx-wiki.atlassian.net/browse/MCKIN-3637) and [OSPR-970](https://openedx.atlassian.net/browse/OSPR-970)
**Testing instructions:** 
No user facing parts - running unittests should be enough.

For manual verification:
1. Create an XBlock with any field that have `default=UNIQUE_ID` and `force_export=True` (or use existing one, i.e. https://github.com/open-craft/xblock-group-project-v2). 
2. Create a course, add that XBlock to course content, *do not specify a value for that field - leave it default*, but note the value somehow (i.e. display it in the student view)
3. Export course.
4. Import course. Observe the value was preserved.

### DIFF
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -480,7 +480,7 @@ class XmlSerializationMixin(ScopedStorageMixin):
         for field_name, field in self.fields.iteritems():
             if field_name in ('children', 'parent', 'content'):
                 continue
-            if field.is_set_on(self) or field.runtime_options.get('force_export', False):
+            if field.is_set_on(self) or field.force_export:
                 self._add_field(node, field_name, field)
 
         # Add children for each of our children.
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -477,7 +477,7 @@ class XmlSerializationMixin(ScopedStorageMixin):
         for field_name, field in self.fields.iteritems():
             if field_name in ('children', 'parent', 'content'):
                 continue
-            if field.is_set_on(self):
+            if field.is_set_on(self) or field.runtime_options.get('force_export', False):
                 self._add_field(node, field_name, field)
 
         # Add children for each of our children.
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -8,6 +8,7 @@ import inspect
 import logging
 from lxml import etree
 import copy
+from collections import OrderedDict
 
 try:
     import simplesjson as json  # pylint: disable=F0401
@@ -22,10 +23,12 @@ from xblock.fields import Field, Reference, Scope, ReferenceList
 from xblock.internal import class_lazy, NamedAttributesMetaclass
 
 
-XML_NAMESPACES = {
-    "option": "http://code.edx.org/xblock/option",
-    "block": "http://code.edx.org/xblock/block",
-}
+# OrderedDict is used so that namespace attributes are put in predictable order
+# This allows for simple string equality assertions in tests and have no other effects
+XML_NAMESPACES = OrderedDict([
+    ("option", "http://code.edx.org/xblock/option"),
+    ("block", "http://code.edx.org/xblock/block"),
+])
 
 
 class HandlersMixin(object):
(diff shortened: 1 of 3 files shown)


edx-webhook:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=326&repo=edx%2FXBlock

edx-webhook:
Thanks for the pull request, @e-kolpakov! I've created [OSPR-970](https://openedx.atlassian.net/browse/OSPR-970) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

Kelketek:
:+1:

xblock/mixins.py
```diff
@@ -22,10 +23,12 @@
...
-XML_NAMESPACES = {
-    "option": "http://code.edx.org/xblock/option",
-    "block": "http://code.edx.org/xblock/block",
-}
+# OrderedDict is used so that namespace attributes are put in predictable order
+# This allows for simple string equality assertions in tests and have no other effects
+XML_NAMESPACES = OrderedDict([
+    ("option", "http://code.edx.org/xblock/option"),
+    ("block", "http://code.edx.org/xblock/block"),
+])
```
doctoryes: I don't see anything added in this PR that requires the `OrderedDict`. What tests require this ordering?

doctoryes:
While the addition of the `force_export` parameter seems fine (and seems needed in this situation), I'd like for the parameter and a description of its use to be added to the docstring here:

https://github.com/edx/XBlock/blob/master/xblock/fields.py#L275-L302

especially since it's directly related to the `default` parameter.

xblock/mixins.py
e-kolpakov: @doctoryes [this test|https://github.com/edx/XBlock/blob/master/xblock/test/test_parsing.py#L299-L317] is flaky because namespaces *might* be put in different order.

e-kolpakov:
@doctoryes since docstring contains constructor parameter definitions, I've converted `force_export` from runtime option to field attribute. And added it to docstring of course.

xblock/mixins.py
doctoryes: Makes perfect sense, thanks.

doctoryes:
:+1:  This PR is ready to merge. I'll do so now.

doctoryes:
@e-kolpakov I assume there'll be a follow-on PR for edx-platform to use this new XBlock feature?

@pdesjardins FYI - might be XBlock documentation to add here.

e-kolpakov:
@doctoryes so far the only use case for this is in an XBlock that runs on edx-solutions fork. So there's no need to change edx/edx-platform requirements right now. But if it is a good moment to update it I can do so. I'm in UTC+7 timezone now, so we're 12 hours apart (if you're in Boston), so I'll wait for your response on this until tomorrow morning my time. If there aren't any - I'll do a hash update PR in edx/edx-platform

doctoryes:
@e-kolpakov It's preferable to go ahead and expose the new functionality now to edx-platform. The alternative would be at some point in the future, someone adds newer XBlock functionality, updates edx-platform, and releases this functionality unknowingly with their own. I don't foresee problems with this XBlock change - but as a general practice, it's better to release it now.

PR #327, opened 2015-11-22
Making filesystem part of core XBlocks

pmitros:
This PR makes pyfilesystem *field* support into a supported feature.

It explicitly does *not* mark the associated XBlock *service* API production-quality.

It is a minimal PR. It adds djpyfs as a dependency. It removes code for handling when it is not there. It removes comments marking this as prototype. It slightly shifts things around in the file to move the service as not finalized, but the field as finalized.

@cpennington @nedbat @ormsbee Would any two of you care to give this a thumbs up?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,11 @@ setup(
         'python-dateutil',
         'pytz',
         'webob',
+        'fs',
     ],
+    extras_require={
+        'django': ['django-pyfs']
+    },
     license='Apache 2.0',
     classifiers=(
         "License :: OSI Approved :: Apache Software License 2.0",
--- a/xblock/reference/plugins.py
+++ b/xblock/reference/plugins.py
@@ -8,23 +8,37 @@ Much of this still needs to be organized.
 """
 
 try:
+    from django.core.exceptions import ImproperlyConfigured
+except ImportError:
+    class ImproperlyConfigured(Exception):
+        '''
+        If Django is installed, and djpyfs is installed, but we're not in a
+        Django app, we'll get this exception. We'd like to catch
+        it. But we don't want the try/except to fail even if we're
+        either in a proper Django app, or don't have Django installed
+        at all.
+        '''
+        pass
+
+try:
     from djpyfs import djpyfs  # pylint: disable=import-error
 except ImportError:
+    djpyfs = None  # pylint: disable=invalid-name
+except ImproperlyConfigured:
+    print "Warning! Django is not correctly configured."
     djpyfs = None  # pylint: disable=invalid-name
 
 from xblock.fields import Field, NO_CACHE_VALUE
 from xblock.fields import scope_key
 
 #  Finished services
-#    None yet
-#
-#  edX-internal prototype services
 
 
 def public(type=None, **kwargs):  # pylint: disable=unused-argument, redefined-builtin
     """
-    Mark a function as public. In the future, this will inform the XBlocks services
-    framework to make the function remotable. For now, this is a placeholder.
+    Mark a function as public. In the future, this will inform the
+    XBlocks services framework to make the function remotable. For
+    now, this is a placeholder.
 
     The kwargs will contain:
 
@@ -53,21 +67,23 @@ def public(type=None, **kwargs):  # pylint: disable=unused-argument, redefined-b
 class Service(object):
     """
     Top-level definition for an XBlocks service.
-    This is intended as a starting point for discussion, not a finished interface.
+
+    This is intended as a starting point for discussion, not
+    necessarily a finished interface.
 
     Possible goals:
-
-    * Right now, they derive from object. We'd like there to be a common superclass.
-    * We'd like to be able to provide both language-level and service-level bindings.
-    * We'd like them to have a basic knowledge of context (what block they're being
-      called from, access to the runtime, dependencies, etc.
-    * That said, we'd like to not over-initialize. Services may have expensive
-      initializations, and a per-block initialization may be prohibitive.
-    * We'd like them to be able to load through Stevedor, and have a plug-in
-      mechanism similar to XBlock.
-
-    This superclass should go somewhere else. This is an interrim location until we
-    figure out where.
+    * Right now, they derive from object. We'd like there to be a
+      common superclass.
+    * We'd like to be able to provide both language-level and
+      service-level bindings.
+    * We'd like them to have a basic knowledge of context (what block
+      they're being called from, access to the runtime, dependencies,
+      etc.
+    * That said, we'd like to not over-initialize. Services may have
+      expensive initializations, and a per-block initialization may be
+      prohibitive.
+    * We'd like them to be able to load through Stevedor, and have a
+      plug-in mechanism similar to XBlock.
     """
     def __init__(self, **kwargs):
         # TODO: We need plumbing to set these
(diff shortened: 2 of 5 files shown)


pmitros:
Test pass. The previous test failure was unrelated to the PR. XBlock has a number of flaky tests. In this case, the ordering of XML attributes was random, and the test had an exact string compare for import/export. One was 

     <leafwithoption xmlns:block="http://code.edx.org/xblock/block" xmlns:option="http://code.edx.org/xblock/option"

The other was: 

     <leafwithoption xmlns:option="http://code.edx.org/xblock/option" xmlns:block="http://code.edx.org/xblock/block

cpennington:
We talked in the past about making it so that (say) django specific functionality would be available if you have django, but wouldn't otherwise cause import errors (and making the dependency easy for runtimes to declare by using an `extras_require` in setup.py). Now seems like an ideal time to do this, but it would mean separating the service (which currently requires django) from the field (which could be implemented without django).

pmitros:
Good catch. I'll poke at that.

pmitros:
@cpennington Like this? (note: still need to test that all works with xblock-sdk; I just ran XBlock tests alone)

setup.py
```diff
@@ -24,6 +24,10 @@
         'python-dateutil',
         'pytz',
         'webob',
+        'fs',
+    ],
+    extra_requires=[
+        'django-pyfs',
```
cpennington: Your syntax here isn't quite right. Check out https://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

Lyla-Fischer:
Is there any documentation of the discussion that lead to the decision of making the filesystem part of core XBlocks? The discussion on these threads ([thread 1](https://groups.google.com/forum/#!topic/edx-code/aofXeyHmaYs) [thread 2](https://groups.google.com/forum/#!msg/edx-code/4VadWwqeMNI/K33-ilJXH1kJ)) seems to indicate that there might be performance issues associated with the particular library (djpyfs) that this pull request includes as a dependency. 

Furthermore, an instance of djpyfs is returned in FSService with minimal wrapping, meaning that this pull request is tightly coupling core XBlock with django with minimal additional abstraction. It might be better for XBlocks that need a filesystem to make their own design choices in which filesystem to select.

xblock/reference/plugins.py
```diff
@@ -171,3 +130,42 @@ def __set__(self, xblock, value):
         We don't support this until we figure out what this means
```
Lyla-Fischer: Is there documentation for that this means, now that it is going to be supported?

pmitros: `__set__` is not supported, as indicated by the comment.

pmitros:
We discussed at arch lunch. It's been in production for an extended period without performance issues. @nedbat @cpennington and @ormsbee were all on board with promoting to supported. 

We do not couple to `djpyfs`. That's a reference implementation. We couple to `pyfilesystem` which is a standard Python package. We add two operations to it: 

1. A way to expire objects
2. A way to return a URL for an object with a time limit 

Any runtime can do the same for any backend implementation.

xblock/reference/plugins.py
Lyla-Fischer: It was my understanding that the point of this pull request was to move the filesystem field into supported status. I'm confused how that can be done when the __set__ method for the field remains unsupported.

pmitros: StackOverflow has several good articles explaining the `__set__` method and there are several nice howtos on Python descriptors.

Lyla-Fischer: I'm sorry, I should have been more clear. I was not intending to ask about python syntax. I was hoping that you could replace the comments about prototypes with a description of exactly what is supported about this field. It looks like you are supporting a read-only file system, and the applications of that are not obvious.

pmitros: You were abundantly clear. I did my best to point you to documentation which would answer your question, which was (as your clarification again shows) based on a misunderstanding of Python descriptors. It is not a write-only file system. You interact with a file system by opening files and writing to them, not by assigning to the file system itself. It is not at all clear what assignment to a file system would mean, so that operation is unsupported.

I would suggest looking over the code of file thumbs or the Recommender, as you had previously indicated you might do. Both write to the file system. I would also look at pyfilesystem documentation.

Lyla-Fischer: I guess that it just seems weird to me to move a field to "supported" when so many of the interactions with the field will be using dependencies that are implemented in an unsupported prototype. I'm not sure what the guarantees are that the "supported" aspect of the system is trying to provide, other than that an unsupported system also exists.

pmitros: In this case, the distinction is simple: 

* We are moving the field type to be a supported, stable API. People can rely on this API to continue to exist.
* We are not moving the service to be a supported, stable API. That means we may change it in the future if we come up with a better way to do things. If we do change this, the field type will continue to work, since we control both ends. 
* Both are currently considered to have robust, stable implementations. It's just that we are only committing to one of them as a long-term contract. If XBlocks use the fs field type, they will continue to work for whatever support timeframe we end up having on the current API, and with similar deprecation mechansims. If XBlocks use the service, they may break at any time as we evolve the API. 

Again, it would be helpful if you actually took the time to read the references I already gave you several times. At the very least, you should take the time to actually understand the code *before* commenting further. The PR mechanism is intended as a way to comment on specific code, not to request a free programming tutorial.

nedbat: @pmitros The comment "until we figure out what this means" could certainly use clarification.  Also `__set__`, while standard Python, is one of its most advanced and obscure features. It seems very reasonable to me to ask for more explanation here.

setup.py
```diff
@@ -24,7 +24,11 @@
         'python-dateutil',
         'pytz',
         'webob',
+        'fs',
     ],
+    extras_require={
+        'django-fs-support': ['django-pyfs']
```
nedbat: Who is going to use this extras declaration?

xblock/reference/plugins.py
```diff
@@ -171,3 +130,42 @@ def __set__(self, xblock, value):
...
+    def load(self, instance, xblock):
+        """
+        Get the filesystem for the field specified in 'instance' and the
+        xblock in 'xblock' It is locally scoped.
+        """
+
+        # TODO: Get xblock from context, once the plumbing is piped through
+        try:
+            return djpyfs.get_filesystem(scope_key(instance, xblock))
+        except NameError:
```
nedbat: The line of code here could raise NameError for a number of reasons.  I'd rather keep the logic at the top of the file that sets djpyfs to None if it can't be imported, and then check it for None here.

nedbat: You can check for None inside the method, rather than as a way to conditionally define the class.

xblock/test/test_parsing.py
```diff
@@ -19,7 +19,7 @@
 
 def get_namespace_attrs():
     """ Returns string suitable to be used as an xmlns parameters in XBlock XML representation """
-    return " ".join('xmlns:{}="{}"'.format(k, v) for k, v in XML_NAMESPACES.items())
+    return " ".join('xmlns:{}="{}"'.format(k, v) for k, v in sorted(XML_NAMESPACES.items()))
```
nedbat: Why was this needed?

pmitros: An unrelated test was intermittently failing. It was parsing and generating XML, comparing exact strings. This makes XML generation deterministic.

cpennington: I like consistently generated XML. Another option, though, would be to compare parsed XML objects.

xblock/reference/plugins.py
cpennington: I agree, I think explaining that `__set__` isn't a sensible operation on a field which represents an entire filesystem would be reasonable.

setup.py
cpennington: A runtime that uses django, and wants the provided django-backed filesystem.

Although, I might just shorten it to `'django': ['django-pyfs']` and plan to include all django-specific dependencies in it.

xblock/test/test_parsing.py
pmitros: @cpennington @nedbat See the current test failure to see the test flakiness here. This was passing prior to my current commit (which is unrelated to this test).

I am en route to SFO right now, so I don't have a great way to test locally (I pushed and relied on Travis), but I believe I addressed all issues? 

I'm not 100% certain about optional requires -- the way pip is handling it on my system is completely broken, and if one of you were to confirm it works, I'd be grateful.

pmitros:
@nedbat @cpennington Could I get a thumbs up please? I rebased and fixed conflicts to merge to master and removed code which was already redone upstream (e.g. fixing the flakey test). 

While I was at it, I did a few doc cleanups as well. 

Also, what's the deal with coveralls? That's shown as a failing check, but I'd expect 0.09% changes in coverage either way to be reasonable to merge with.... Is that a new requirement, or is that informational? It seems helpful as an informational bit, but a little weird to show as a test failure.

cpennington:
Yeah, the coveralls thing is annoying. I think if we got https://github.com/edx/XBlock/pull/296 merged, then we'd actually be able to test both the with and without scenarios, but I wouldn't want to block this PR on that.

cpennington:
:+1:

PR #332, opened 2016-02-17
Support XML serialization of None for String field.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-736

See related PR: https://github.com/edx/edx-platform/pull/11490

Add ability to send a String field value of `None` to the XBlock's XML representation using a XML attribute on a XML element. Deserialize the `None` value upon loading from XML. Add more testing around String fields with different defaults.

Also, fix a bug with caching the default value after deleting an XBlock field.

@cpennington @adampalay Please review. Anyone else that should review?
@macdiesel FYI.

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -307,6 +307,8 @@ class Field(Nameable):
     """
     MUTABLE = True
     _default = None
+    # Indicates if a field's None value should be sent to the XML representation.
+    none_to_xml = False
 
     # We're OK redefining built-in `help`
     def __init__(self, help=None, default=UNSET, scope=Scope.content,  # pylint:disable=redefined-builtin
@@ -464,6 +466,19 @@ class Field(Nameable):
         key = scope_key(self, xblock)
         return hashlib.sha1(key).hexdigest()
 
+    def _get_default_value_to_cache(self, xblock):
+        """
+        Perform special logic to provide a field's default value for caching.
+        """
+        try:
+            # pylint: disable=protected-access
+            return self.from_json(xblock._field_data.default(xblock, self.name))
+        except KeyError:
+            if self._default is UNIQUE_ID:
+                return self._check_or_enforce_type(self._calculate_unique_id(xblock))
+            else:
+                return self.default
+
     def __get__(self, xblock, xblock_class):
         """
         Gets the value of this xblock. Prioritizes the cached value over
@@ -482,13 +497,7 @@ class Field(Nameable):
                 value = self.from_json(field_data.get(xblock, self.name))
             elif self.name not in NO_GENERATED_DEFAULTS:
                 # Cache default value
-                try:
-                    value = self.from_json(field_data.default(xblock, self.name))
-                except KeyError:
-                    if self._default is UNIQUE_ID:
-                        value = self._check_or_enforce_type(self._calculate_unique_id(xblock))
-                    else:
-                        value = self.default
+                value = self._get_default_value_to_cache(xblock)
             else:
                 value = self.default
             self._set_cached_value(xblock, value)
@@ -549,7 +558,7 @@ class Field(Nameable):
         # Since we know that the field_data no longer contains the value, we can
         # avoid the possible database lookup that a future get() call would
         # entail by setting the cached value now to its default value.
-        self._set_cached_value(xblock, copy.deepcopy(self.default))
+        self._set_cached_value(xblock, self._get_default_value_to_cache(xblock))
 
     def __repr__(self):
         return "<{0.__class__.__name__} {0.name}>".format(self)
@@ -828,6 +837,11 @@ class String(JSONField):
     def to_string(self, value):
         """String gets serialized and deserialized without quote marks."""
         return self.to_json(value)
+
+    @property
+    def none_to_xml(self):
+        """Returns True to use a XML node for the field and represent None as an attribute."""
+        return True
 
     enforce_type = from_json
 
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -449,13 +449,13 @@ class XmlSerializationMixin(ScopedStorageMixin):
             namespace = qname.namespace
 
             if namespace == XML_NAMESPACES["option"]:
-                cls._set_field_if_present(block, tag, child.text)
+                cls._set_field_if_present(block, tag, child.text, child.attrib)
             else:
                 block.runtime.add_node_as_child(block, child, id_generator)
 
         # Attributes become fields.
         for name, value in node.items():
-            cls._set_field_if_present(block, name, value)
+            cls._set_field_if_present(block, name, value, {})
 
         # Text content becomes "content", if such a field exists.
         if "content" in block.fields and block.fields["content"].scope == Scope.content:
@@ -507,32 +507,43 @@ class XmlSerializationMixin(ScopedStorageMixin):
             return None
 
     @classmethod
-    def _set_field_if_present(cls, block, name, value):
+    def _set_field_if_present(cls, block, name, value, attrs):
         """Sets the field block.name, if block have such a field."""
         if name in block.fields:
             value = (block.fields[name]).from_string(value)
-            setattr(block, name, value)
+            if "none" in attrs and attrs["none"] == "true":
+                setattr(block, name, None)
+            else:
+                setattr(block, name, value)
         else:
             logging.warn("XBlock %s does not contain field %s", type(block), name)
 
     def _add_field(self, node, field_name, field):
-        """Add xml representation of field to node.
+        """
+        Add xml representation of field to node.
 
         Depending on settings, it either stores the value of field
         as an xml attribute or creates a separate child node.
         """
         value = field.to_string(field.read_from(self))
+        text_value = "" if value is None else value
 
-        if value is None:
-            value = ""
+        # Is the field type supposed to serialize the fact that the value is None to XML?
+        save_none_as_xml_attr = field.none_to_xml and value is None
+        field_attrs = {"none": "true"} if save_none_as_xml_attr else {}
 
-        if field.xml_node:
+        if save_none_as_xml_attr or field.xml_node:
+            # Field will be output to XML as an separate element.
             tag = etree.QName(XML_NAMESPACES["option"], field_name)
-            elem = node.makeelement(tag)
-            elem.text = value
-            node.insert(0, elem)
+            elem = etree.SubElement(node, tag, field_attrs)
+            if field.xml_node:
+                # Only set the value if forced via xml_node;
+                # in all other cases, the value is None.
+                # Avoids an unnecessary XML end tag.
+                elem.text = text_value
         else:
-            node.set(field_name, value)
+            # Field will be output to XML as an attribute on the node.
+            node.set(field_name, text_value)
 
 
 class IndexInfoMixin(object):
--- a/xblock/test/test_mixins.py
+++ b/xblock/test/test_mixins.py
@@ -11,7 +11,7 @@ from unittest import TestCase
 from xblock.core import XBlock
 from xblock.fields import List, Scope, Integer, String, ScopeIds, UNIQUE_ID, DateTime
 from xblock.field_data import DictFieldData
-from xblock.mixins import ScopedStorageMixin, HierarchyMixin, IndexInfoMixin, ViewsMixin
+from xblock.mixins import ScopedStorageMixin, HierarchyMixin, IndexInfoMixin, ViewsMixin, XML_NAMESPACES
 from xblock.runtime import Runtime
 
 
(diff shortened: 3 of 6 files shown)


xblock/fields.py
```diff
@@ -652,6 +659,12 @@ def is_set_on(self, xblock):
         # pylint: disable=protected-access
         return self._is_dirty(xblock) or xblock._field_data.has(xblock, self.name)
 
+    def none_to_xml(self):
```
cpennington: Maybe just make this a class attribute to start with, and then if a field needs it to be instance specific they can override w/ a property?

doctoryes: Sure - I'll make the change.

xblock/fields.py
```diff
@@ -549,7 +558,7 @@ def __delete__(self, xblock):
         # Since we know that the field_data no longer contains the value, we can
         # avoid the possible database lookup that a future get() call would
         # entail by setting the cached value now to its default value.
-        self._set_cached_value(xblock, copy.deepcopy(self.default))
+        self._set_cached_value(xblock, self._get_default_value_to_cache(xblock))
```
doctoryes: @cpennington I was seeing a failure in a test that had a String field with a default of `UNIQUE_ID` when the field was being set and then deleted. This change fixed the failure - but is it correct?

xblock/fields.py
doctoryes: Done.

xblock/fields.py
```diff
@@ -829,6 +838,11 @@ def to_string(self, value):
         """String gets serialized and deserialized without quote marks."""
         return self.to_json(value)
 
+    @property
+    def none_to_xml(self):
+        """Returns True to use a XML node for the field and represent None as an attribute."""
+        return True
```
adampalay: Why is this being set on a property instead of class level, when it's set on a class level in `Field`? Curious :)

xblock/mixins.py
```diff
@@ -507,32 +507,44 @@ def xml_text_content(self):
...
         Depending on settings, it either stores the value of field
         as an xml attribute or creates a separate child node.
         """
         value = field.to_string(field.read_from(self))
+        text_value = "" if value is None else value
 
-        if value is None:
-            value = ""
+        # Is the field type supposed to serialize the fact that the value is None to XML?
+        none_to_xml_attr = field.none_to_xml and value is None
```
adampalay: maybe better name is something like `should_serialize_none_to_xml`?

xblock/mixins.py
```diff
@@ -507,32 +507,44 @@ def xml_text_content(self):
...
+        field_attrs = {"none": "true"} if none_to_xml_attr else {}
 
-        if field.xml_node:
+        if none_to_xml_attr or field.xml_node:
+            # Field will be output to XML as an separate element.
             tag = etree.QName(XML_NAMESPACES["option"], field_name)
-            elem = node.makeelement(tag)
-            elem.text = value
+            elem = node.makeelement(tag, field_attrs)
```
adampalay: python's xml module suggests using `etree.SubElement` instead: https://docs.python.org/2/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.makeelement

xblock/test/test_mixins.py
```diff
@@ -260,38 +271,151 @@ def _assert_node_attributes(self, node, expected_attributes):
...
+    def _assert_node_elements(self, node, expected_elements, check_none_attr=True):
+        """
+        Checks XML node elements to match expected elements.
+        """
+        node_elements = list(node)
+        self.assertEqual(set([elem.tag for elem in node_elements]), set(expected_elements.keys()))
+        if check_none_attr:
+            # All elements on the node are expected to have a "none"="true" attribute.
+            for elem in node:
+                self.assertEqual(elem.get('none'), 'true')
```
adampalay: should we assert that the `node` has elements? (just to make sure we hit L283?)

xblock/test/test_mixins.py
```diff
@@ -260,38 +271,151 @@ def _assert_node_attributes(self, node, expected_attributes):
...
+                'str_str_default_force_export': 'default',
+                'str_uid_default_force_export': UNIQUE_ID
+            }
         )
+        self._assert_node_elements(
+            node,
+            {
+                # The tags are prefixed with {namespace}.
+                '{{{}}}{}'.format(XML_NAMESPACES["option"], tag): None
+                for tag in ['str_none_default_force_export']
```
adampalay: why is this a comprehension if there's only one tag we're iterating over?

xblock/test/test_mixins.py
```diff
@@ -260,38 +271,151 @@ def _assert_node_attributes(self, node, expected_attributes):
...
+        Tests that set fields appear in XML after add_xml_to_node.
+        """
+        node = etree.Element(self.test_xblock_tag)
 
-        block.field = 'Value1'
-        block.simple = 'Value2'
-        block.simple_force_export = 'Value3'
-        block.unique_id = 'Value4'
-        block.unique_id_force_export = 'Value5'
+        self.test_xblock.str_field = 'Value1'
```
adampalay: Could you make these values more descriptive of what they are? Like `"string_field"` instead of `"Value1"`?

xblock/fields.py
doctoryes: Using a `@property` decorator in this class makes the attribute immutable for this class. So no code can change a String field to *not* send Nones to XML. It's not that important - just an extra safeguard.

adampalay: oh nice, that's a really good call

xblock/mixins.py
doctoryes: How about `save_none_as_xml_attr`? Short for "Save a None value as an XML attribute on an XML element."

adampalay: :+1:

xblock/test/test_mixins.py
doctoryes: Good call - will do.

doctoryes: Actually - there's a case where no nodes are expected. But I did check to make sure the line is being called in other cases (manual check).

xblock/test/test_mixins.py
doctoryes: I was following the pattern used elsewhere. But you're right - I'll simplify it.

xblock/test/test_mixins.py
doctoryes: Will do.

xblock/mixins.py
doctoryes: Great - I'll use that.

adampalay:
Hey @doctoryes , I have a question about the [`XML_NAMESPACE` OrderedDict](https://github.com/edx/XBlock/blob/master/xblock/mixins.py#L26-L32). The namespaces refer to URLs that no longer exist: both 
http://code.edx.org/xblock/option
and
http://code.edx.org/xblock/block
404. Should these be updated? The comment that precedes its assignment says that it's only used in tests, but I can imagine more industrious course teams potentially being confused by xml that looks like:

```xml
<library xblock-family="xblock.v1" text_customization="{}" rerandomize="never" graded="false" tags="[]" edxnotes="false" annotation_storage_url="http://your_annotation_storage.com" use_latex_compiler="false" video_bumper="{}" course_edit_method="Studio" static_asset_path="" hide_from_toc="false" in_entrance_exam="false" showanswer="finished" display_name="test" group_access="{}" video_speed_optimizations="true" graceperiod="null" annotation_token_secret="xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" due="null" user_partitions="[]" start="&quot;2030-01-01T00:00:00Z&quot;" show_reset_button="false" days_early_for_beta="null" max_attempts="null" visible_to_staff_only="false" org="test" library="test">
  <ns8:chrome xmlns:ns8="http://code.edx.org/xblock/option" none="true"/>
  <ns7:name xmlns:ns7="http://code.edx.org/xblock/option" none="true"/>
  <ns6:default_tab xmlns:ns6="http://code.edx.org/xblock/option" none="true"/>
  <ns5:source_file xmlns:ns5="http://code.edx.org/xblock/option" none="true"/>
  <ns4:giturl xmlns:ns4="http://code.edx.org/xblock/option" none="true"/>
  <ns3:format xmlns:ns3="http://code.edx.org/xblock/option" none="true"/>
  <ns2:xqa_key xmlns:ns2="http://code.edx.org/xblock/option" none="true"/>
  <ns1:matlab_api_key xmlns:ns1="http://code.edx.org/xblock/option" none="true"/>
  <ns0:advanced_modules xmlns:ns0="http://code.edx.org/xblock/option">[]</ns0:advanced_modules>
  <html url_name="f54a1fd18cc4401eb0e78bbf11b6e454"/>
</library>
```

Thoughts?

cpennington:
@adampalay generally speaking, XML namespaces aren't required (and often aren't) actually urls that have content.

doctoryes:
It's intentional that those XML namespaces don't yet exist. The comment to which you refer talks about the order in the OrderedDict being used by a test. But the actual namespaces are used in code to differentiate XBlock fields as XML nodes. Those namespaces were added in this PR:

https://github.com/edx/XBlock/pull/244

But they've never been defined.

Is it the length of the XML that's confusing? Because technically there's no need for 9 different namespaces in your example. The XML could (possibly) instead look like this:
```
<library 
xmlns:xbo="http://code.edx.org/xblock/option" xblock-family="xblock.v1" text_customization="{}" rerandomize="never" graded="false" tags="[]" edxnotes="false" annotation_storage_url="http://your_annotation_storage.com" use_latex_compiler="false" video_bumper="{}" course_edit_method="Studio" static_asset_path="" hide_from_toc="false" in_entrance_exam="false" showanswer="finished" display_name="test" group_access="{}" video_speed_optimizations="true" graceperiod="null" annotation_token_secret="xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" due="null" user_partitions="[]" start="&quot;2030-01-01T00:00:00Z&quot;" show_reset_button="false" days_early_for_beta="null" max_attempts="null" visible_to_staff_only="false" org="test" library="test">
  <xbo:chrome none="true"/>
  <xbo:name none="true"/>
  <xbo:default_tab none="true"/>
  <xbo:source_file none="true"/>
  <xbo:giturl none="true"/>
  <xbo:format none="true"/>
  <xbo:xqa_key none="true"/>
  <xbo:matlab_api_key none="true"/>
  <xbo:advanced_modules >[]</xbo:advanced_modules>
  <html url_name="f54a1fd18cc4401eb0e78bbf11b6e454"/>
</library>
```
It's a bit strange that `advanced_modules` ended up as an element instead of attribute above - since it's a List and not a String.

adampalay:
Ah @cpennington , thanks for that clarifying context! It was just confusing to me because I didn't know those urls weren't expected to resolve

adampalay:
Yeah, the advanged_modules thing was a surprise too. Any idea why that happened?

doctoryes:
@adampalay Ah - mystery solved. In a content library, that field is forced to be an XML node:

https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/library_root_xblock.py#L35

adampalay:
Wow, that was an easy resolution :)

:+1: from me. This is great, thanks

xblock/fields.py
cpennington: Not sure... What behavior was the test expecting? What was it getting?

xblock/test/test_mixins.py
cpennington: It seems like maybe there should be a separate `assert_node_nones` (or something) that does the assertion that's currently happening if `check_none_attr` is true. It seems odd to have an assertion function that takes an argument to do an optional additional assertion.

cpennington: Oh. I guess `check_none_attr` is never `False`... Maybe just remove that argument?

cpennington:
Couple of questions, but otherwise :+1:

xblock/fields.py
doctoryes: @cpennington IIRC, the test was failing due trying to deepcopy the `UNIQUE_ID` Sentinel. The `__get__` method above was using different logic to retrieve the default value, so I split that logic out and used it here as well. The test expected to default the cached value back to the same Sentinel default.

xblock/test/test_mixins.py
doctoryes: Yes - I agree. I'll remove it.

xblock/fields.py
cpennington: Hm. The intention was that `Sentinels` would be singletons, so they should probably behave correctly under deepcopy (and just return themselves, so that `is` checks work).

adampalay:
:fireworks:

PR #335, opened 2016-03-16
WL-376 move i18n runtime logic in parent XBlock

asadiqbal08:
@mattdrayer @nedbat please review the change set.

### DIFF
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -174,6 +174,15 @@ class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeS
         """
         return Validation(self.scope_ids.usage_id)
 
+    def ugettext(self, text):
+        """
+        Translates message/text and returns it in a unicode string.
+        Using runtime to get i18n service.
+        """
+        runtime_service = self.runtime.service(self, "i18n")
+        runtime_ugettext = runtime_service.ugettext
+        return runtime_ugettext(text)
+
 
 class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixin, HandlersMixin, SharedBlockBase):
     """
--- a/xblock/test/test_runtime.py
+++ b/xblock/test/test_runtime.py
@@ -581,6 +581,24 @@ def test_service():
     runtime.render(tester, 'student_view')
 
 
+def test_ugettext_calls():
+    """
+    Test ugettext calls in xblock.
+    """
+    runtime = TestRuntime()
+    block = XBlockWithServices(runtime, scope_ids=Mock(spec=[]))
+    assert_equals(block.ugettext('test'), u'test')
+    assert_true(isinstance(block.ugettext('test'), unicode))
+
+    # NoSuchServiceError exception should raise if i18n is none/empty.
+    runtime = TestRuntime(services={
+        'i18n': None
+    })
+    block = XBlockWithServices(runtime, scope_ids=Mock(spec=[]))
+    with assert_raises(NoSuchServiceError):
+        block.ugettext('test')
+
+
 @XBlock.needs("no_such_service_sub")
 @XBlock.wants("another_not_service_sub")
 class SubXBlockWithServices(XBlockWithServices):


mattdrayer:
Already discussed offline, but just wanted to add a note here that we need to include test coverage with this changeset.

xblock/core.py
```diff
@@ -174,6 +174,15 @@ def validate(self):
...
         return Validation(self.scope_ids.usage_id)
 
+    def _(self, text):
+        """
+        Hand the translation work off to the runtime's I18N service
+        """
+        runtime_service = self.runtime.service(self, "i18n")
+        runtime_ugettext = runtime_service.ugettext
+        response = runtime_ugettext(text)
+        return response
```
nedbat: Let's save the name "response" for an HttpResponse.

xblock/core.py
```diff
@@ -174,6 +174,15 @@ def validate(self):
         """
         return Validation(self.scope_ids.usage_id)
 
+    def _(self, text):
```
nedbat: I'm wondering if we should use ugettext for this.  Also, please write a docstring that describes what this function actually does.

asadiqbal08:
Updated the PR as per suggestions. Please look inside.

mattdrayer:
I'm trying to determine if the test failure @ Travis is due to our changes -- it seems unrelated, but perhaps  the charset is being included now due to the I18N support?

asadiqbal08:
@mattdrayer I was trying to produce this error locally but running tests locally are passing.

mattdrayer:
@nedbat and....@doctoryes?  The test failure we are observing with this PR is also occurring with a simple test PR (#336) -- thoughts?

doctoryes:
@asadiqbal08 @mattdrayer Looks like the test passes using Django 1.8.9 - and fails using Django 1.8.11. I assume that Django 1.8.10 would fail also. I'd advise looking over the 1.8.10 release notes to see what changed enough to cause the test failure.

xblock/core.py
```diff
@@ -174,6 +174,15 @@ def validate(self):
         """
         return Validation(self.scope_ids.usage_id)
 
+    def ugettext(self, text):
```
asadiqbal08: @mattdrayer what you suggest if we keep the name of this function as `_` instead of `ugettext` ?

mattdrayer: Hmm, not sure I understand -- I think we should keep the name of this function "ugettext" as advised by @nedbat.  In the individual XBlocks we might think about setting a module-level variable "_" as a reference to "self.ugettext" for marking up the actual strings -- is that what you mean?

asadiqbal08:
@mattdrayer @doctoryes I was looking around the test failure issue. It seems me , `webob version` is causing the failure.
travis has `webob=1.6.0` while on my local environment I have `webob=1.5.1`. 
Kindly look into the green test build [#336](https://github.com/edx/XBlock/pull/336)

asadiqbal08:
@mattdrayer FYI [look here](https://github.com/Pylons/webob/pull/197/files#diff-ecaf7f8adaceba1e9c89f4db7d179f6cR131)

mattdrayer:
@nedbat @doctoryes I'd like to merge this PR today and continue to focus on #338 as a separate issue -- please let me know if this does not work for you.

xblock/core.py
cpennington: Why are we adding this particular shortcut to the base XBlock? Just so that we can call `self.ugettext(...)` easily, rather than needing to explicitly say `self.runtime.service(self, 'i18n').ugettext(...)`?

mattdrayer: Yes, that's the basic idea -- generalizing a common pattern that we're seeing across multiple XBlock projects

mattdrayer:
Alright, PR #338 has been updated to fix the WebOb test failure and merged to master -- I've rebased this branch and it appears we now have a green build -- let me know if you guys would like to see anything else -- if not would be great to see a couple thumbs so we can move ahead with the merge :smile:

xblock/test/test_runtime.py
```diff
@@ -581,6 +581,24 @@ def test_service():
...
 
 
+def test_ugettext_calls():
+    """
+    Test ugettext calls in xblock.
+    """
+    runtime = TestRuntime()
+    block = XBlockWithServices(runtime, scope_ids=Mock(spec=[]))
+    assert_equals(block.ugettext('test'), u'test')
+    isinstance(block.ugettext('test'), unicode)
```
doctoryes: This `isinstance` call needs an associated `assert` call.

mattdrayer: I've added an assert_true assertion

doctoryes:
:+1:

mattdrayer:
Nice!  Merging now -- thank you!

PR #336, opened 2016-03-17
testing PR

asadiqbal08:
just travis testing. I will remove that PR

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 pyyaml
 lxml
 requests
-webob
+webob == 1.5.1
 simplejson
 pytz
 python-dateutil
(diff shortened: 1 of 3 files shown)


mattdrayer:
Hmm, that is interesting -- I know the XBlock project has been around for a while, but I thought we were generally trying to avoid tracking latest for dependencies, and instead pinning to specific versions.  @doctoryes are you the right person to be involved in working through this?  @nedbat?

I'm inclined to let #335 go ahead seeing that the Travis failure is unrelated to the work we've done -- we can either log a ticket to address the webob issue (to which project -- Platform?), or we could go ahead and pin it now as a separate PR -- what would you guys prefer?

mattdrayer:
@nedbat @doctoryes, just bumping this one up on your lists

doctoryes:
My choice is to fix up this PR which pins `webob` to 1.5.1 - and merge it before merging #335 .
:+1:

feanil:
Removing this Test PR from 3 years ago.

PR #337, opened 2016-03-22
Fix XBlockAside.add_xml_to_node crash

StraNNiKK:
This PR is related with the conversation: https://github.com/edx/edx-platform/pull/11808#discussion_r56330814

`has_children` attribute is set in [ChildrenModelMetaclass](https://github.com/edx/XBlock/blob/master/xblock/mixins.py#L333) metaclass. `ChildrenModelMetaclass` is metaclass for [HierarchyMixin](https://github.com/edx/XBlock/blob/master/xblock/mixins.py#L355). `XBlock` inherits `HierarchyMixin`:
```
class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin,
RuntimeServicesMixin, HandlersMixin, IndexInfoMixin, ViewsMixin, SharedBlockBase):
```
but `XBlockAside` isn't:
```
class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, 
RuntimeServicesMixin, HandlersMixin, SharedBlockBase):
```
The `self.has_children` check is used in `XmlSerializationMixin.add_xml_to_node` function.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,3 +36,4 @@ Dhruv Baldawa <dhruvbaldawa@gmail.com>
 Matjaz Gregoric <mtyaka@gmail.com>
 Adam Palay <adam@edx.org>
 Awais Jibran <awaisdar001@gmail.com>
+Dmitry Viskov <dmitry.viskov@webenterprise.ru>
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -183,6 +183,14 @@ class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeS
         runtime_ugettext = runtime_service.ugettext
         return runtime_ugettext(text)
 
+    def add_xml_to_node(self, node):
+        """
+        For exporting, set data on etree.Element `node`.
+        """
+        super(XBlock, self).add_xml_to_node(node)
+        # Add children for each of our children.
+        self.add_children_to_node(node)
+
 
 class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixin, HandlersMixin, SharedBlockBase):
     """
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -413,6 +413,15 @@ class HierarchyMixin(ScopedStorageMixin):
         """
         self._child_cache.clear()
 
+    def add_children_to_node(self, node):
+        """
+        Add children to etree.Element `node`.
+        """
+        if self.has_children:
+            for child_id in self.children:
+                child = self.runtime.get_block(child_id)
+                self.runtime.add_block_as_child_node(child, node)
+
 
 class XmlSerializationMixin(ScopedStorageMixin):
     """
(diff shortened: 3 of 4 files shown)


edx-webhook:
Thanks for the pull request, @StraNNiKK! I've created [OSPR-1200](https://openedx.atlassian.net/browse/OSPR-1200) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

cpennington:
:+1:

cpennington:
Thanks for fixing this!

jbarciauskas:
@cpennington do we just need to press the green button?

xblock/test/django/test_request.py
```diff
@@ -80,12 +80,12 @@ def test_headers(self):
     def test_content_type(self):
         # Default charset
         self.assertEquals(
-            self._as_django(content_type='application/json')['Content-Type'],
-            'application/json; charset=UTF-8'
+            self._as_django(content_type='text/html')['Content-Type'],
+            'text/html; charset=UTF-8'
```
nedbat: I'm concerned about these test changes.  It seems to be that we are moving from a test that behaves similar to real code, to a test that behaves less like real code.  PR #338 has much more discussion about it.

StraNNiKK: @nedbat ok. Do I need to revert these changes?

mattdrayer:
Okay, so based on the context in #338, where are we at now with this PR?  Sorry to be such a pain about it, but I really would like to see #335 merge soon.

mattdrayer:
Checking back in on this PR -- where are we at with it?

cpennington:
@mattdrayer how is this related to #335?

cpennington:
I thought that this was targeted at fixing a bug found in https://github.com/edx/edx-platform/pull/11808

xblock/test/django/test_request.py
cpennington: @nedbat: All this test was intending to test was the behavior of the `charset` argument to `_as_django`. I think the main issue with the test as it was written as that it assumed it knew what charset webob was going to provide by default, and thus it broke when webob changed that default. I think the test should just be changed to assert that the content-type is set to the supplied content type, but ignore the charset, and then if the charset is set (as below), verify that `_as_django` sets the charset correctly.

mattdrayer:
I was told to hold off on merging #335 until #337 merged, because it fixes the issue being discussed in #338 (although this PR doesn't appear to address the problem correctly).  IMHO neither #337 nor #338 actually block #335, so I would like to just merge it, but I'm trying to work with peoples' concerns.

cpennington:
Yeah, #335 seems unrelated to the charset issues to me, except that its suffering from breakage from an unpinned webob version. If we want to have tests pass, then pinning webob to the current version (that the code is known to work with) in #335, and then upgrading webob in a separate PR is probably the way to go (at least in my mind).

mattdrayer:
If we *must* see green on #335 then I am happy to pin to webob v1.5.1 so we can move forward.  We had done this at one point but removed the pin after some discussion.

nedbat:
Is there a reason it's hard to fix #338 the right way and do the right thing everywhere?

mattdrayer:
I'm still not clear on what the 'right way' is?

mattdrayer:
OKAY, after discussing further with @nedbat we determined what needed to be done with PR #338 and that branch has been updated and merged -- a rebase is now in order for this change set, which should address the WebOb-related test failure.

StraNNiKK:
@mattdrayer @nedbat @cpennington hi. I've made rebase and updated the PR

jbarciauskas:
@mattdrayer @nedbat @cpennington is this ready to go?

PR #338, opened 2016-03-22
webob version

asadiqbal08:
@mattdrayer @nedbat @doctoryes please review it. 

I make this change in order resolve the test failure issue in #335

### DIFF
--- a/xblock/test/django/test_request.py
+++ b/xblock/test/django/test_request.py
@@ -77,15 +77,15 @@ class TestDjangoWebobResponse(TestCase):
         self.assertIn('X-Foo', self._as_django(headerlist=[('X-Foo', 'bar')]))
         self.assertEquals(self._as_django(headerlist=[('X-Foo', 'bar')])['X-Foo'], 'bar')
 
-    def test_content_type(self):
-        # Default charset
+    def test_content_types(self):
+        # JSON content type (no charset should be returned)
         self.assertEquals(
             self._as_django(content_type='application/json')['Content-Type'],
-            'application/json; charset=UTF-8'
+            'application/json'
         )
 
-        # Other charset
+        # HTML content type (UTF-8 charset should be returned)
         self.assertEquals(
-            self._as_django(content_type='application/json', charset='ascii')['Content-Type'],
-            'application/json; charset=ascii'
+            self._as_django(content_type='text/html')['Content-Type'],
+            'text/html; charset=UTF-8'
         )
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 pyyaml
 lxml
 requests
-webob
+webob>=1.6.0
 simplejson
 pytz
 python-dateutil


mattdrayer:
:+1:

doctoryes:
:+1:

asadiqbal08:
@nedbat ^^ or go with merge  ?

nedbat:
We've had a theory that tight pins like this in libraries make it difficult to combine libraries together.  is there a reason a range of versions won't work?

mattdrayer:
We could probably do a range here, as long as it is something like <1.6 which is where we are seeing the problem -- any recommendations?

nedbat:
I don't have the context of what the problem was.  Pinning to an older than current is tricky.  Have we asked them to fix the bug they introduced?

nedbat:
It looks to me like 1.6.0 didn't introduce a bug, it fixed one: application/json shouldn't specify a charset.  Why don't we fix the test, and pin to >=1.6.0 ?

doctoryes:
I actually agree with Ned here - I have thumb remorse. Fixing the test will avoid the whole issue.

mattdrayer:
Okay, we will update the branch to reflect this feedback and hopefully will be looking good for a merge tomorrow.

asadiqbal08:
@nedbat just for your info, I had shared some of my findings here #335 , Actually `charset` will not be the part of response If we have `aplication/json` as you can see [here](https://github.com/Pylons/webob/pull/197/files#diff-ecaf7f8adaceba1e9c89f4db7d179f6cR131) for more details. 
Ok I am going to update the test.

asadiqbal08:
@nedbat @mattdrayer please look inside.

xblock/test/django/test_request.py
```diff
@@ -80,12 +80,12 @@ def test_headers(self):
     def test_content_type(self):
         # Default charset
         self.assertEquals(
-            self._as_django(content_type='application/json')['Content-Type'],
-            'application/json; charset=UTF-8'
+            self._as_django(content_type='text/plain')['Content-Type'],
+            'text/plain; charset=UTF-8'
```
nedbat: We don't make Django responses that are plain text, we make JSON responses, so I think it's important to keep the same content_type.

xblock/test/django/test_request.py
```diff
@@ -80,12 +80,12 @@ def test_headers(self):
     def test_content_type(self):
         # Default charset
         self.assertEquals(
-            self._as_django(content_type='application/json')['Content-Type'],
+            self._as_django(content_type='application/json; charset=UTF-8')['Content-Type'],
```
asadiqbal08: @nedbat thoughts ?

mattdrayer: Looks right to me?

cpennington:
Is there an api that we can use to check/set the charset that isn't changed between webob versions? I don't think we care about the charset explicitly, so I'd rather have our tests just not break whether or not it adds the charset.

xblock/test/django/test_request.py
nedbat: Honestly, I'm having a hard time deciding on the right answer here. BUT: webob seems to have decided that character sets are wrong in a json content type.  The important thing is that these tests call this function in a way similar to how it is called in edx-platform.  I don't know what that way is, but this fix is both changing the way the function is called in the test, and avoiding the change that webob thought was important.  This worries me.

mattdrayer:
I'm happy to continue working on this PR, but I'm getting concerned that it is blocking #335 and doesn't seem to be related -- can we let that one fly?

doctoryes:
@mattdrayer Let's wait until https://github.com/edx/XBlock/pull/337 gets merged before merging https://github.com/edx/XBlock/pull/335. It fixes the webob test and is close to being merged.

mattdrayer:
I have to admit I'm fairly confused by this whole webob broken test situation.
If #337 fixes the problem, that's fine, but I don't see how what is being done there is different from what was proposed in https://github.com/edx/XBlock/pull/338/commits/43ff928b6b70e67bda70449bc03cba9704a359be, which we decided we didn't like.
Also in #337 we are not updating the actual webob reference to reflect the range as we did in this changeset.

TBH my priority is #335 so if #337 is good to go then I'll just let it be and move on :)

doctoryes:
@mattdrayer Sorry - I mis-read the other PR, thinking that @cpennington had approved the test fix when the commit was actually added after his thumb.

@nedbat and I discussed this fix - the best fix (for now) seems to be pinning the webob version to >=1.6.0 and removing the `charset` expectation from the test. I don't understand completely why webob decided to remove the `charset` - that's the remaining question for me.

xblock/test/django/test_request.py
```diff
@@ -80,12 +80,12 @@ def test_headers(self):
...
             'application/json; charset=UTF-8'
         )
 
         # Other charset
         self.assertEquals(
-            self._as_django(content_type='application/json', charset='ascii')['Content-Type'],
+            self._as_django(content_type='application/json; charset=ascii')['Content-Type'],
```
bertjwregeer: An ASCII charset for JSON is always invalid, and is not allowed.

bertjwregeer:
We have removed the `charset` from JSON responses because according to the JSON spec, the valid way to determine the charset is by using magic byte detection for UTF-8, UTF-16 or UTF-32 (first four bytes determine charset), also the media type as registered for JSON does not specify a charset:

http://www.iana.org/assignments/media-types/application/json

Currently there are still ways that WebOb will spit out the `charset=utf-8` (such as for instance in WebOb exceptions that are JSON'ified, which I need to get fixed), but a charset specified in the `content-type` is contra-spec.

This was first noted in Pyramid by @dstufft while he was working on Warehouse: https://github.com/Pylons/pyramid/issues/1611.

`application/json` is the only valid `content-type` for JSON, and the only valid encodings are UTF-{8,16,32}.

mattdrayer:
Thanks @bertjwregeer !!!

xblock/test/django/test_request.py
cpennington: `_as_django` takes the charset and content type separately so that the caller doesn't have to worry about the correct format of the header. Can we leave the signature as it was?

Also, as I mentioned here: https://github.com/edx/XBlock/pull/337/files, I think that this test should just validate that specifying `content_type` gives you the content type you asked for (without asserting anything about the charset), and that setting both gives you a header that has both in it.

nedbat:
:+1:

mattdrayer:
Thanks much @nedbat -- going ahead with the merge now

PR #340, opened 2016-04-14
Add the ability to control the root where resources load from.

tobz:
This feature lets an XBlock configure, precisely, where its public assets actually live on disk.  This is useful for a few reason:
- packages which have multiple XBlocks can separate the public folders for each XBlock to ensure isolation
- XBlocks can set the value to None which prevents them from having local_resource_url invoked on them period, which is useful for signalling that an XBlock should not be trying to even use resources

This is important for a few XBlock authors but also to allow an easier migration of XModules/XBlocks in the edx-platform to become freestanding instead of globbed together.

### DIFF
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -50,6 +50,9 @@ class SharedBlockBase(Plugin):
     """
     Behaviors and attrs which all XBlock like things should share
     """
+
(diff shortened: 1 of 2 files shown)


cpennington:
:+1:

PR #344, opened 2016-04-22
Handle the resources directory better.

tobz:
The first pass at making the`resources_dir` configurable allowed XModules and XBlocks bundled in **edx-platform** to work, but broke an implicit assumption by standalone XBlocks that you could access resource URLs rooted at the XBlock root, or more precisely, the root of the source control repository.  This would be the standard "public" folder that XModules and XBlocks use already, and they would reference `public/js/foo.js` in their code.

The original change, coupled with the XBlock asset pipeline work in **edx-platform**, broke this by collecting assets in a way where the leading directory was chopped off.  Asking for `public/js/foo.js` would no longer work, because an XBlock only saw itself as having `js/foo.js` available to it.

This work adds back the concept of what the public folder should be.  In conjunction with setting the resources directory, an XModule/XBlock can both set a non-default resources root -- important for when multiple XModules or XBlocks are present in the same source control repository/folder, and also customize which folder represnets their "public" resources, or resources that can be presented externally by the LMS.


xblock/core.py
```diff
@@ -51,15 +52,27 @@ class SharedBlockBase(Plugin):
...
-    resources_dir = 'public'
+    resources_dir = ''
+    public_dir = 'public'
 
     @classmethod
     def get_resources_dir(cls):
+        """
+        Gets the resource directory for this XBlock.
+        """
         return cls.resources_dir
```
cpennington: Sigh. I feel like we should have just had this as a class attr that could have been made a property later... oh well.

cpennington:
:+1:

PR #346, opened 2016-04-22
Add a way to stream unprotected local resources.

tobz:
Depending on whether or not you're in a production environment, you may want to be able to stream all resources from a given XBlock.  This happens, in particular, when edx-platform is running in debug mode, and all resources are required individually, instead of being bundled up.

Instead of requiring consumers to do their own pkg_resources lookups to sidestep open_local_resource's 'public' restriction, we now have a method that allows this to happen in a controlled way from the XBlock itself, which takes into consideration things like a custom resources_dir configuration, etc.

### DIFF
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -103,6 +103,16 @@ class SharedBlockBase(Plugin):
         if "/." in uri:
             raise DisallowedFileError("Only safe file names are allowed: %r" % uri)
 
+        return cls.stream_local_resource(uri)
+
+    @classmethod
+    def stream_local_resource(cls, uri):
+        """
+        Stream a local resources.
+
+        This is a helper method to get a file-like handle to a local resource
+        while adjusting URIs based on the resource configuration of this XBlock.
+        """
         return pkg_resources.resource_stream(cls.__module__, os.path.join(cls.resources_dir, uri))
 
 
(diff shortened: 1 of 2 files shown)


tobz:
Example of what this is intended to solve/prevent:

    if settings.DEBUG:
        content = pkg_resources.resource_stream(xblock_class.__module__, uri) <----
    else:
        content = xblock_class.open_local_resource(uri)

(https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/common_views/xblock.py#L29)

cpennington:
Why do we need to bypass the `public` check? Shouldn't the blocks act the same in debug mode? As it stands, when debugging, you might not notice that your resources are in the wrong spot.

tobz:
This is for cases, at least per the comments in `edx-platform`, where things might not be bundled yet into the larger bundles we generate.  In that case, resources would be getting requested individually, and they might not live in the `public` folder.

cpennington:
Aren't the files at least in the base folder that the XBlock specified, though?

tobz:
Not sure.  I was simply mimicing the existing behavior.

cpennington:
It seems like, since you've made the blocks configurable now (so that they don't *require* the data to be in public), that you likely don't need this exception anymore.

tobz:
Forgot to close this out, but: yeah, seems like you are totally right here.

Issue #334, opened 2016-03-05
Typo `simplesjson` -> `simplejson`


abduld:
there are two occurrences of this mistake in the code

~~~
try:
    import simplesjson as json  # pylint: disable=F0401
except ImportError:
    import json
~~~

nedbat:
Fixed in #347, thanks.

PR #347, opened 2016-05-05
There's no such module as simplesjson  #334

nedbat:


nedbat:
@cpennington buh?

cpennington:
:+1:

... argh ...

PR #349, opened 2016-06-16
Introduce python 3 compatibility.

jcdyer:
## Hackathon project (no JIRA ticket)

This ticket introduces python 3 support to the XBlock repo, while maintaining backwards compatibility with python 2 using future and six.  
### Reviewers:
- [ ] @nedbat 
- [x] @cpennington 
### Notes:
- I opted to use and `six.text_type`, as opposed to `future.builtins.str`, as per discussions of the python 3 working group.
- I opted to use `from __future__ import unicode_literals` throughout rather than explicit `u'unicode'` string literals.
- I haven't cleaned up all the explicit unicode literals.
- I have tried to make the distinction between binary and text explicit in the tests without changing the supported behavior.  However:
- lxml is a little tricky with python 3, because if you include the character set in the `<?xml?>` declaration, the input has to be a bytestring of that encoding, and if you want to output a character set declaration, you have to output bytes in that character set.  This makes sense, in some ways, but it also conflates the creation of a serialized xml string with the encoding of that string for output, which makes it difficult to create a proper unicode sandwich with our current interface (which passes around xml as strings rather than as etree documents).  I've thus left the parsing code as returning bytes, and accepting either strings or bytes, but subject to lxml's finickiness.
- I had to upgrade pylint and astroid to get them to work with python3.  This has caused numerous new pylint errors.  I've cleaned most of them up.
- Tests now run through tox, and run against both python 2 and python 3.


xblock/fields.py
```diff
@@ -779,6 +779,18 @@ def from_json(self, value):
...
+        """
+        In python3, json.dumps() cannot sort keys of different types, 
+        so preconvert None to 'null'.
+        """
+        self.enforce_type(value)
+        if isinstance(value, dict) and None in value:
+            value = value.copy()
+            value['null'] = value[None]
+            del value[None]
+        return super(Dict, self).to_string(value)
```
jcdyer: @cpennington We have a test that checks that things work with null keys.  Do we also allow numeric keys?  Because those would also fail for the same reason (different types not comparable in python3), and we might need to do the same kind of processing for them here.  

However, with None, we can perform the fix in constant time, whereas for numbers, we would have to process each key in turn, so I didn't want to do the work unless there's a use for it.

cpennington: Hrm... we don't explicitly disallow numeric keys on the python side, so for completeness, we probably should be checking for numeric keys. This is all making dropping the `sort_keys` parameter sound better to me, just from the perspective of limiting the amount of preprocessing we have to do over and above what the json module already does.

cpennington: I'm not sure if you found http://bugs.python.org/issue25457, but perhaps referencing it when removing `sort_keys`.

jcdyer: I hadn't seen that.  The pushback against fixing it is not surprising, but it is frustrating.  You shouldn't be able to break json.dumps just by specifying a cosmetic option. :(

One easyish option that I've played with is to leave in the sorted version, but catch the TypeError, and retry with sort_keys=False.

jcdyer:
Note: I updated pylint, because the specified version 1.3.1 was hilariously broken under python 3 (e.g. `undefined variable 'self'`, where self is obviously defined), but this has introduced a number of new linting errors due to changed pylint expectations.

xblock/django/request.py
```diff
@@ -1,9 +1,11 @@
 """Helpers for WebOb requests and responses."""
+from builtins import object
```
jcdyer: I don't love that future monkeypatches builtins.  I'm tempted to use the more explicit `from future.builtins import object`.  This `object` lets you use python 3 dunder-methods in python 2 (like __next__, __str__ returning unicode, etc).

xblock/field_data.py
```diff
@@ -88,7 +91,7 @@ def set_many(self, block, update_dict):
         :param update_dict: A map of field names to their new values
         :type update_dict: dict
         """
-        for key, value in update_dict.items():
+        for key, value in list(update_dict.items()):
```
jcdyer: I just noticed all of these, from the autoconversion.  I will update them to use `six.iteritems()` (and related functions) to avoid the extra memory usage.

jcdyer:
See also edx-platform integration branch at https://github.com/edx/edx-platform/pull/13263

cpennington:
Which interfaces pass around XML strings? I thought all of the XML parsing interfaces just used etree objects.

xblock/core.py
```diff
@@ -90,6 +99,9 @@ def open_local_resource(cls, uri):
         unauthorized resource.
         """
 
+        if isinstance(uri, six.binary_type):
+            uri = uri.decode('utf-8')
```
cpennington: I'd rather have our methods be specific about what type of string they are going to take, rather than allowing both types.

Per https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax (because wikipedia is the authoritative source of everything), only ASCII characters are allowed.

xblock/fields.py
```diff
@@ -752,7 +766,9 @@ def __init__(self, help=None, default=UNSET, scope=Scope.content, display_name=N
                                       **kwargs)
 
     def from_json(self, value):
-        if isinstance(value, basestring):
+        if isinstance(value, six.binary_type):
+            value = value.decode('latin-1')  # guaranteed successful decode
```
cpennington: Why is `latin-1` guaranteed to decode? What if value were encoded as something else?

cpennington: Again, I think I'd rather be strict about string types in these methods, if possible.

cpennington: Although, I suppose this method used to not be strict...

xblock/core.py
jcdyer: I would also prefer to be strict, but the existing implementation was flexible, so maintaining backwards compatibility requires less than optimal unicode handling.  We could issue warnings when binary objects are passed in.

jcdyer: Good point, though about URIs.  I'll change the decode to ascii only.

xblock/fields.py
jcdyer: latin-1 is guaranteed to decode, because every possible byte value can be represented as a unicode value.  If the value were encoded as something else, it would decode as a different codepoint than intended, but it would not lose any data.  In this instance, that doesn't matter, because for any codepoint outside of ascii, `value.lower()` will not equal `'true'`.  

I was just made aware yesterday of the `errors` argument to `bytes.decode()`, which would make the intention clearer: `value.decode('ascii', errors='replace')`.  I will update this.

xblock/fields.py
```diff
@@ -839,18 +867,18 @@ def _sanitize(self, value):
...
         Leave all other characters.
         """
-        if isinstance(value, unicode):
-            new_value = u''.join(ch for ch in value if unicodedata.category(ch)[0] != u'C' or ch in (u'\n', u'\r', u'\t'))
-        elif isinstance(value, str):
-            new_value = ''.join(ch for ch in value if ord(ch) >= 32 or ch in ('\n', '\r', '\t'))
+        if isinstance(value, six.text_type):
+            new_value = u''.join(ch for ch in value if unicodedata.category(ch)[0] != u'C' or ch in u'\n\r\t')
+        elif isinstance(value, six.binary_type):
+            new_value = ''.join(ch for ch in value if ord(ch) >= 32 or ch in b'\n\r\t')
```
jcdyer: I suspect it would be better to handle this by converting the binary value to text first, and only stripping bad characters once.

xblock/run_script.py
```diff
@@ -13,6 +20,6 @@ def run_script(pycode):
 
     # execute it.
     globs = {}
-    exec pycode in globs, globs  # pylint: disable=W0122
+    exec(pycode, globs, globs)  # pylint: disable=W0122
```
jcdyer: Hmm.  This must be untested.  I think this would break in python2.  It should be using the six version of exec.

jcdyer:
@cpennington The XML handling does use lxml etree objects internally, but it has to parse in data at somepoint, and output data at another point.  

`self.runtime.export_to_xml(block, output)` is where it outputs bytes.  

`self.runtime.parse_xml_string(xml)` is where we take in either binary or string data.  If the xml starts with an `<?xml?>` declaration, lxml requires it to be a bytes object of the declared encoding.  If the declaration is missing, it requires a unicode object.

cpennington:
Ah, that makes sense.

xblock/core.py
jcdyer: Hmm.  We actually test that utf-8 URLs work, using a file called `u"public/ℓιвяαяу.js"` as an example.

jcdyer: I think the catch here is that the variable doesn't describe a URI but a *local* resource.

jcdyer:
@cpennington @nedbat @maxrothman @jmbowman This is up-to-date with the latest version of xblock.  Looking for another round of review, if anyone's interested.  

Merging this is still blocked by coveralls not reporting.

edx-platform tests also pass against this version of xblock.  See: edx/edx-platform#13263

Makefile
```diff
@@ -1,12 +1,7 @@
-test:
-	coverage run -m nose
-
-docs:
```
nedbat: It'd be cool to keep these targets, with "tox -e docs" as the action.

xblock/__init__.py
```diff
@@ -4,6 +4,11 @@
 
 # For backwards compatability, provide the XBlockMixin in xblock.fields
 # without causing a circular import
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
```
nedbat: Putting these four on one line will fit into our line-length limit.

xblock/django/request.py
```diff
@@ -20,7 +27,7 @@ def webob_to_django_response(webob_response):
     return django_response
 
 
-class HeaderDict(MutableMapping):
+class HeaderDict(MutableMapping, six.Iterator):
```
nedbat: I haven't seen this before: we do we need to add Iterator?

xblock/plugin.py
```diff
@@ -173,7 +178,7 @@ def test_the_thing():
         def _decorator(func):                           # pylint: disable=C0111
             @functools.wraps(func)
             def _inner(*args, **kwargs):                # pylint: disable=C0111
-                global PLUGIN_CACHE
+                global PLUGIN_CACHE  # pylint: disable=global-statement
```
nedbat: Micro-nit: I prefer to tab the pragmas out to the right so they are less distracting.

xblock/runtime.py
```diff
@@ -1086,7 +1092,7 @@ class BadPath(Exception):
             """Bad path exception thrown when path cannot be found."""
             pass
         results = self.query(block)
-        ROOT, SEP, WORD, FINAL = range(4)               # pylint: disable=C0103
+        ROOT, SEP, WORD, FINAL = six.moves.range(4)               # pylint: disable=C0103
```
nedbat: This one we could have left alone :)

xblock/runtime.py
```diff
@@ -1278,6 +1284,31 @@ def strftime(self, dtime, format):      # pylint: disable=redefined-builtin
...
+    @property
+    def ugettext(self):
+        """
+        Dispatch to the appropriate gettext method
+        """
+        # pylint: disable=no-member
+        if six.PY3:
+            return self._translations.gettext
+        else:
+            return self._translations.ugettext
```
nedbat: I'm a bit concerned that the underlying functions change names on 2 vs 3, but our wrapper does not?  At least the docstring should make clear what this function will do.

xblock/runtime.py
```diff
@@ -1278,6 +1284,31 @@ def strftime(self, dtime, format):      # pylint: disable=redefined-builtin
...
+
+    @property
+    def ungettext(self):
```
nedbat: Ditto, of course.

xblock/test/test_core.py
```diff
@@ -939,48 +950,78 @@ def stub_resource_stream(self, module, name):
...
 
     @ddt.data(
-        "public/../secret.js",
-        "public/.git/secret.js",
-        "static/secret.js",
-        "../public/no-no.bad",
-        "image.png",
-        ".git/secret.js",
-        "static/ℓιвяαяу.js",
+        u"public/hey.js".encode('utf-8'),
```
nedbat: Why not use `b""` strings here?

xblock/test/test_parsing.py
```diff
@@ -123,24 +129,24 @@ class ParsingTest(XmlTest, unittest.TestCase):
 
     @XBlock.register_temp_plugin(Leaf)
     def test_parsing(self):
-        block = self.parse_xml_to_block("<leaf data2='parsed'/>")
+        block = self.parse_xml_to_block(u"<leaf data2='parsed'/>")
```
nedbat: No need to add these u-prefixes, right? unicode_literals will take care of it.

xblock/test/test_parsing.py
```diff
@@ -360,19 +383,27 @@ def create_block(self, block_type):
...
-        expected_dict = {'1': '1', 'ping': 'ack'}
+        expected_seq = [b'1', b'2']
+        expected_dict = {b'1': b'1', b'ping': b'ack'}
         block.sequence = expected_seq
         block.dictionary = expected_dict
-        xml = self.export_xml_for_block(block)
-
-        parsed = self.parse_xml_to_block(xml)
+        if six.PY3:
+            self.assertRaises(TypeError, self.export_xml_for_block, block)
```
nedbat: Why does this raise an exception on Py 3.

xblock/django/request.py
jcdyer: It lets you write a class with `__next__()`, and have it be an iterator on python 2 (having a `next()` method.  But on python3 it doesn't add the `next()` method. https://pythonhosted.org/six/#six.Iterator

xblock/test/test_core.py
jcdyer: In part because we would have had to represent the multi-byte values as their individual bytes.  I don't remember if there was a reason to do them for the ascii-equivalent bytestrings.

xblock/test/test_parsing.py
jcdyer: Yep.  I thought I had cleaned it all up.  I'll make another pass to see what else I missed.

xblock/test/test_core.py
nedbat: Hmm, I see. I never tried using non-ASCII in bytestrings on py3: you can't! :)

tox.ini
```diff
@@ -0,0 +1,27 @@
...
+# in multiple virtualenvs. This configuration file will run the
+# test suite on all supported python versions. To use it, "pip install tox"
+# and then run "tox" from this directory.
+
+[tox]
+envlist = py27, py35, docs
+
+[testenv]
+commands =
+    coverage run -m nose
```
nedbat: This results in a .coverage data file in the project root, which gets overwritten between py27 and py35. Looking at some of these tricky conditionals, I've wondered about their coverage, and I don't have a good way to see.

tox.ini
```diff
@@ -0,0 +1,27 @@
...
+    make quality
+deps =
+    -rrequirements.txt
+    -rdjango_requirements.txt
+    codecov
```
nedbat: Is this used somewhere? I really should know more about how codecov works....

tox.ini
cpennington: We should make sure to tag the uploads to codecov with the version of python that they were generated on.

cpennington: http://docs.codecov.io/docs/flags is the docs.

xblock/runtime.py
jcdyer: True.

xblock/test/test_parsing.py
jcdyer: See the docstring above:
```
        """
        Test correctly serializes-deserializes List and Dicts with byte string
        contents in Python 2.

        In Python 3, this behavior is unsupported.  dict and list elements
        cannot be bytes objects.
        """
```

tox.ini
jcdyer: I'll make that change.

xblock/runtime.py
maxrothman: Also, didn't we recommend to check for py2 and have the else hit py3 for future compatibility with py4?

Makefile
jcdyer: Done.

xblock/__init__.py
jcdyer: Done.

xblock/runtime.py
jcdyer: I added comments to the docstrings for these functions.  I agree that the naming mismatch is not ideal, but I think it's valuable to have the discrepancy wrapped up and kept away from day-to-day development.

jcdyer:
I've tried to address all issues mentioned.  

I'm still not seeing the codecov flags show up in the report, though.  If either of you have suggestions on what I'm doing wrong, or want to add a patch to fix the tox/travis configurations, I'd be grateful.

jcdyer:
@Bednar this is ready for another round of review

jmbowman:
@jcdyer django-pyfs 1.0.5 added Python 3 support last month, might want to set that as a lower bound in django_requirements.txt and/or setup.py.  Other than resolving merge conflicts, does this need anything more than final reviews to move forward?

xblock/internal.py
```diff
@@ -58,8 +63,8 @@ class Nameable(object):
     :class:`.NamedAttributesMetaclass`, will be assigned a `__name__`
     attribute based on what class attribute they are bound to.
     """
-    __slots__ = ('__name__')
-
+    if six.PY2:
+        __slots__ = ('__name__',)
     __name__ = None
```
jcdyer: @cpennington  This breaks under python 3 because slots only apply to instance level attributes, and it doesn't like having a class attribute of the same name.  Python 2 seems okay with it, so I left it conditionally.  Note also that the __slots__ was originally mistakenly defined as a string, rather than as a tuple of strings.  The docs are ambivalent about whether that works.

django_requirements.txt
```diff
@@ -1,4 +1,4 @@
 # Install Django requirements, if we're using the optional Django-integrated
 # parts of XBlock
 Django >= 1.8, < 1.9
-django-pyfs
+django-pyfs >= 1.0.5
```
jcdyer: @jmbowman: This is pinned now.

setup.py
```diff
@@ -33,7 +36,7 @@
         'web-fragments',
     ],
     extras_require={
-        'django': ['django-pyfs']
+        'django': ['django-pyfs >= 1.0.5']
```
jcdyer: @jmbowman Also pinned in setup.py here.

xblock/__init__.py
```diff
@@ -2,8 +2,16 @@
...
+# For backwards compatability, provide the XBlockMixin in xblock.fields
+# without causing a circular import
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import codecs
 import os
 import warnings
+
+import six
```
jcdyer: Yep.  Odd that pylint didn't mind.

requirements.txt
```diff
@@ -27,7 +28,8 @@ ddt==0.8.0
 cookiecutter
 
 # For web fragments
-web-fragments
+#web-fragments
+git+https://github.com/edx/web-fragments.git@0.1.1#egg=web_fragments==0.1.1
```
jcdyer: I don't remember why this was changed.  The pypi version passes tests.

jcdyer:
@jmbowman Responded to all comments.  Ready for another round.

nasthagiri:
Congrats!

PR #351, opened 2016-06-29
Remove control characters in XBlock String fields.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-1031

If a control character exists in an XBlock String field that is then sent to XML, a ValueError is raised. This error blocks a course with control characters in video display_names (for example) from being exported. A simple example of the problem:
```
>>> from lxml import etree
>>> xml = etree.Element('my_element')
>>> string_field = u'Control Characters: "What to do \vabout them?"'
>>> string_field
u'Control Characters: "What to do \x0babout them?"'
>>> xml.set('my_attr', string_field)
  File "<stdin>", line 1, in <module>
  File "lxml.etree.pyx", line 744, in lxml.etree._Element.set (src/lxml/lxml.etree.c:44335)
  File "apihelpers.pxi", line 554, in lxml.etree._setAttributeValue (src/lxml/lxml.etree.c:19531)
  File "apihelpers.pxi", line 1393, in lxml.etree._utf8 (src/lxml/lxml.etree.c:27125)
ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters
```
I've implemented a solution in this PR which removes invalid XML characters during the `__set__` of a String field - and also removes the characters upon a `__get__` of the field. It uses a general `_sanitize` method that any field can implement if needed.

The downsides to this solution:
- The invalid XML characters are removed silently and are a side-effect of setting/getting the field value, which can be confusing to a user.
- There are places in code (tests at least) that check object equality between what is set and gotten later. The code returns the original string if no changes have occurred - but creates/uses a new str/unicode object if it needs to remove characters.

Alternatives:
- Use XML character references to preserve the control characters in an XML export - and then restore the characters upon import. This solution would make the string above be written to XML like this: `u'Control Characters: "What to do &#11;about them?"'`. However, I don't see much value in preserving these unprintable characters in String fields.
- Raise a ValueError upon `__set__` of a String field with control characters - and deal with the existing cases via modifying course data to remove the characters -or- removing them in code during a String field's `__get__` like the code in this PR does.

Accompanying edx-platform PR (to test the change):
https://github.com/edx/edx-platform/pull/12893

@cpennington @jmbowman @adampalay @macdiesel Please review.

### DIFF
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -17,6 +17,7 @@ import traceback
 import warnings
 import json
 import yaml
+import unicodedata
 
 from xblock.internal import Nameable
 
@@ -479,6 +480,13 @@ class Field(Nameable):
             else:
                 return self.default
 
+    def _sanitize(self, value):
+        """
+        Allow the individual fields to sanitize the value being set -or- "get".
+        For example, a String field wants to remove control characters.
+        """
+        return value
+
     def __get__(self, xblock, xblock_class):
         """
         Gets the value of this xblock. Prioritizes the cached value over
@@ -507,7 +515,7 @@ class Field(Nameable):
         if self.MUTABLE:
             self._mark_dirty(xblock, value)
 
-        return value
+        return self._sanitize(value)
 
     def __set__(self, xblock, value):
         """
@@ -520,6 +528,7 @@ class Field(Nameable):
         we're trying to cache, we won't do anything.
         """
         value = self._check_or_enforce_type(value)
+        value = self._sanitize(value)
         cached_value = self._get_cached_value(xblock)
         try:
             value_has_changed = cached_value != value
@@ -824,9 +833,25 @@ class String(JSONField):
     """
     MUTABLE = False
 
+    def _sanitize(self, value):
+        """
+        Remove the control characters that are not allowed in XML:
+        https://www.w3.org/TR/xml/#charsets
+        Leave all other characters.
+        """
+        if isinstance(value, unicode):
+            new_value = u''.join(ch for ch in value if unicodedata.category(ch)[0] != u'C' or ch in (u'\n', u'\r', u'\t'))
+        elif isinstance(value, str):
+            new_value = ''.join(ch for ch in value if ord(ch) >= 32 or ch in ('\n', '\r', '\t'))
+        else:
+            return value
+        # The new string will be equivalent to the original string if no control characters are present.
+        # If equivalent, return the original string - some tests check for object equality instead of string equality.
+        return value if value == new_value else new_value
+
     def from_json(self, value):
         if value is None or isinstance(value, basestring):
-            return value
+            return self._sanitize(value)
         else:
             raise TypeError('Value stored in a String must be None or a string, found %s' % type(value))
 
--- a/xblock/test/test_fields.py
+++ b/xblock/test/test_fields.py
@@ -69,7 +69,7 @@ class FieldTest(unittest.TestCase):
             if issubclass(warning.category, DeprecationWarning)
         ))
 
-    def assertJSONOrSetEquals(self, expected, arg):
+    def assertJSONOrSetGetEquals(self, expected, arg):
         """
         Asserts the result of field.from_json and of setting field.
         """
@@ -77,6 +77,12 @@ class FieldTest(unittest.TestCase):
         self.assertEqual(expected, self.FIELD_TO_TEST().from_json(arg))
         # set+get with enforce_type arg -> expected
         self.assertEqual(expected, self.set_and_get_field(arg, True))
+
+    def assertJSONOrSetEquals(self, expected, arg):
+        """
+        Asserts the result of field.from_json and of setting field.
+        """
+        self.assertJSONOrSetGetEquals(expected, arg)
         # set+get without enforce_type arg -> arg
         # provoking a warning unless arg == expected
         count = 0 if arg == expected else 1
@@ -227,6 +233,15 @@ class StringTest(FieldTest):
         self.assertJSONOrSetTypeError([1])
         self.assertJSONOrSetTypeError([])
         self.assertJSONOrSetTypeError({})
+
+    def test_control_characters_filtered(self):
+        self.assertJSONOrSetGetEquals(u'', u'\v')
+        self.assertJSONOrSetGetEquals('', '\v')
+        with self.assertRaises(AssertionError):
+            self.assertJSONOrSetGetEquals('\v', u'')
+        with self.assertRaises(AssertionError):
+            self.assertJSONOrSetGetEquals(u'\v', '')
+        self.assertJSONOrSetGetEquals(u'\n\r\t', u'\n\v\r\b\t')
 
 
 @ddt.ddt


xblock/fields.py
```diff
@@ -824,9 +833,23 @@ class String(JSONField):
...
+        https://www.w3.org/TR/xml/#charsets
+        Leave all other characters.
+        """
+        if isinstance(value, unicode):
+            new_value = u''.join(ch for ch in value if unicodedata.category(ch)[0] != u'C' or ch in (u'\n', u'\r', u'\t'))
+        elif isinstance(value, str):
+            new_value = ''.join(ch for ch in value if ord(ch) >= 32 or ch in ('\n', '\r', '\t'))
+        else:
+            return value
+        return value if value == new_value else new_value
```
adampalay: would it be more readable if we did something like: 

```python
    else:
        new_value = value
    return new_value
```

adampalay:
one small comment, o/w 👍

cpennington:
To me, escaping the characters going into the XML seems like the right call, rather than silently dropping them. Whether that happens at the field boundary or at the XML boundary is an open question, although I think I'd lean towards pushing it towards the XML boundary (basically, stick with the "escape for the appropriate context" mantra)

doctoryes:
@cpennington I've looked into escaping the characters going into the XML more. It looks like we'd need a separate, OLX-specific escaping step before sending String fields to XML - and then another separate, OLX-specific escaping step from XML in order to properly preserve the control characters. This separate escaping step on both sides would make OLX less like standard XML, which doesn't support these control characters. Here's some code demonstrating:
```
>>> from lxml import etree
>>> xml = etree.Element('my_element')
>>> string_field = u'Control Characters: "What to do \vabout them?"'
>>> # OLX-specific escaping step here...
...
>>> convert_v1 = 'Control Characters: "What to do &#11;about them?"'
>>> convert_v2 = 'Control Characters: "What to do &#x000B;about them?"'
>>> xml.set('attr1', convert_v1)
>>> xml.set('attr2', convert_v2)
>>> exported_xml = etree.tostring(xml)
>>> exported_xml
'<my_element attr1="Control Characters: &quot;What to do &amp;#11;about them?&quot;" attr2="Control Characters: &quot;What to do &amp;#x000B;about them?&quot;"/>'
>>> imported_xml = etree.fromstring(exported_xml)
>>> imported_xml.get('attr1')
'Control Characters: "What to do &#11;about them?"'
>>> imported_xml.get('attr2')
'Control Characters: "What to do &#x000B;about them?"'
```
In fact, there's no need to use XML character references for the escaped control characters. The escaping can be whatever we choose that can successfully make the round-trip to/from XML.

It seems like the value added by preserving the control characters is not worth making OLX less portable as standard XML.

cpennington:
I'm not sure how it makes it not standard XML, though. It's just a question of encoding values that aren't supported in XML (just like we encode values that aren't supported in other contexts)

doctoryes:
Right - the output would definitely still conform to standard XML.

But, if the intent of OLX is to create a format for passing around course content in XML, this escaping/unescaping of these characters is an implementation detail that'd need to be added to the standard and then implemented in any other systems that consumed OLX - just as the escaping/unescaping would need to be added to edx-platform. If we just got rid of the control characters, a separate escaping/unescaping would be unnecessary and would not need to be part of the (eventual) OLX standard. I'm unconvinced that the value added by bringing along control characters (like BELL or, in the ticket's case, an unneeded Vertical Tab) would be worth that added complexity.

doctoryes:
@nedbat Any thoughts on this issue?

nedbat:
I vote to scrub the bad characters, and find out why people need them.  I suspect they don't.

nedbat:
BTW, it's nuts that XML can't tunnel these Unicode codepoints.

doctoryes:
To be clear, AFAIK nobody has proposed that they need to use control chars or to preserve them across export/import. The control chars in question seemed to end up in the video titles by mistake.
Here's a decent explanation of why these chars aren't legal in XML:
http://stackoverflow.com/questions/404107/why-are-control-characters-illegal-in-xml-1-0
Bottom line: There's no agreed-upon standard about what these chars are/mean/represent across varied systems and XML is all about interoperability.

xblock/fields.py
doctoryes: Done.

doctoryes:
Closing/re-opening to attempt to get my latest commit to show up in the PR.

doctoryes:
@cpennington Further comments?

macdiesel:
Late to the party but I too am I +1 to scrubbing the non allowed characters. If it's not allowed in XML lets not ship it in OLX. 

Also 👍

xblock/fields.py
doctoryes: Actually - the logic needs to stay as it was. When a new string is formed, it may be equivalent to the original string if there are no control characters present. In that case, the original string needs to be returned because some tests check for *object* equality instead of *string* equality. For example:
https://github.com/edx/XBlock/blob/master/xblock/test/test_fields_api.py#L128
@cpennington Is this semantic change going to cause issues anywhere?

cpennington: The requirement for object identity really should just be for mutable fields. It shouldn't matter for immutable ones like strings.

doctoryes: But this PR changes that statement. If a string is stored in a String field and the string has control characters, it *is* mutated to remove the control characters. If it contains no control characters, then it remains unchanged.
I suppose the immutability of a stored string could be preserved if we `_santize`d the string on the way out to XML only. Would you prefer that to changing the immutability of String fields in this edge case?

cpennington: Mutability, in this case, is about whether a pattern like this is possible:

    value = MutableObject()
    xblock.field = value
    value.mutate()
    xblock.field.read_mutation() # what does this show?

For mutable objects (like lists and dicts), it's important to preserve object identity so that mutations that happen outside the field context are still reflected in the field.

For strings, however, the objects themselves are immutable. You can get different strings by calling methods (like `replace`), but those won't have the same object identity.

doctoryes: I see! Seems like this change then doesn't violate the immutability of String field values after all.

doctoryes:
@cpennington Any further concerns with this PR? I'd like to address them all before merging.

xblock/fields.py
cpennington: It's probably worth a comment about why you have the code the way it is (basically, your answer to @adampalay's question).

cpennington:
:+1:

xblock/fields.py
doctoryes: Added.

nedbat:
@ormsbee mentioned that this sanitation code was showing up hot in performance analyses, so I've been poking around to see how we can change it.

I think the sanitation from this pull request that we have now is stricter, and therefore more expensive, than it needs to be.

The XML spec says:
> Consequently, XML processors MUST accept any character in the range specified for Char.
> [2] | Char | ::= | #x9 \| #xA \| #xD \| [#x20-#xD7FF] \| [#xE000-#xFFFD] \| [#x10000-#x10FFFF] | /* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */

It goes on to encourage avoiding other characters also, but an empirical experiment with lxml shows that it accepts all of the characters in the Char grammar rule, so we don't need to avoid those other characters.

<details>
<summary>The lxml experiment</summary>

```
import sys

from lxml import etree
import six

def valid_char(ch):
    xml = etree.Element('my_element')
    try:
        xml.set('my_attr', ch)
    except ValueError:
        return False
    else:
        return True

BAD_RANGES = [[0, 0]]
for c in range(1, sys.maxunicode+1):
    ch = six.unichr(c)
    if not valid_char(ch):
        if c == BAD_RANGES[-1][1] + 1:
            BAD_RANGES[-1][1] = c
        else:
            BAD_RANGES.append([c, c])

print(hex(sys.maxunicode))
print(BAD_RANGES)

BAD_REGEX = u"["
for lo, hi in BAD_RANGES:
    BAD_REGEX += six.unichr(lo)
    if lo != hi:
        if hi > lo + 1:
            BAD_REGEX += u"-"
        BAD_REGEX += six.unichr(hi)
BAD_REGEX += u"]"

print(repr(BAD_REGEX))
```
prints:
```
0x10ffff
[[0, 8], [11, 12], [14, 31], [55296, 57343], [65534, 65535]]
'[\x00-\x08\x0b\x0c\x0e-\x1f\ud800-\udfff\ufffe\uffff]'
```
</details>

Using `'[\x00-\x08\x0b\x0c\x0e-\x1f\ud800-\udfff\ufffe\uffff]'` as a regex to scrub truly invalid characters is much faster (about 20x?)   Any objections to a pull request to make the change?

ormsbee:
@nedbat: Sounds good to me.

nedbat:
#430

PR #353, opened 2016-07-15
Fix simplejson usage and unicode parsing.

doctoryes:
https://openedx.atlassian.net/browse/TNL-5000
Add a test around unicode expectations in handlers.

Fixes changes introduced in this commit:
https://github.com/edx/XBlock/commit/4379a7e7040b51771a20e53d9fbb5472db7c02f3

@nedbat @cahrens @cpennington Please review.
@ssemenova @robrap  FYI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup
 
 setup(
     name='XBlock',
-    version='0.4.11',
+    version='0.4.12',
     description='XBlock Core Library',
     packages=[
         'xblock',
--- a/xblock/mixins.py
+++ b/xblock/mixins.py
@@ -10,10 +10,7 @@ from lxml import etree
 import copy
 from collections import OrderedDict
 
-try:
-    import simplejson as json   # pylint: disable=F0401
-except ImportError:
-    import json
+import json
 import warnings
 
 from webob import Response
--- a/xblock/test/test_core.py
+++ b/xblock/test/test_core.py
@@ -909,6 +909,18 @@ def test_json_handler_return_response():
     assert_equals(response.content_type, "text/plain")
 
 
+def test_json_handler_return_unicode():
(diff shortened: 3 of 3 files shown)


cpennington:
@doctoryes: I'm not following. Why does simplejson lead us into trouble here? The reason for using it was for faster json parsing

cpennington:
(although, admittedly, we weren't using it until the commit you mentioned)

cahrens:
👍 

@cpennington it changed the return type from unicode to basestring (I'm presuming basestring here). Given that we weren't actually using simplejson, the thought on our end is that it is safer to continue with json since it might break other usages besides ORA.

doctoryes:
@cpennington Specifically, because of this:
```
>>> import json
>>> json.loads('{"submission": "testsq"}')
{u'submission': u'testsq'}

>>> import simplejson
>>> simplejson.loads('{"submission": "testsq"}')
{'submission': 'testsq'}
>>> simplejson.loads(u'{"submission": "testsq"}')
{u'submission': u'testsq'}
```
ORA2 expected unicode strings to always be present via the `XBlock.json_handler` and had a failing validation step because the strings became non-JSON.

cahrens:
More details-- here is the validation code that broke in ORA. We have talked about relaxing this validation, but I think reverting XBlock back to json is safer.

````
    for submission_part in submission:
        if type(submission_part) != unicode:
            return False, message

cahrens:
@nedbat @doctoryes and @cpennington we should decide quickly if we are moving forward with this-- either way a repo is going to need to be versioned and edx-platform updated in the release that was cut this morning.

FYI @robrap (release master)

nedbat:
:+1:

ssemenova:
We also need to bump the version number in edx-platform, right?

PR #355, opened 2016-09-09
Python3: Futurize!

cpennington:
@nedbat, @jcdyer, @maxrothman: This converts XBlock to python3 using `future` and `futurize`. It also converts the tests to use tox, because I found maintaining my own virtualenvs to be tedious, and I had a branch that set up tox lying around from 2015.

### DIFF
--- a/script/max_pylint_violations
+++ b/script/max_pylint_violations
@@ -1,5 +1,5 @@
 #!/bin/bash
-DEFAULT_MAX=$15
+DEFAULT_MAX=15
 
 pylint xblock | tee /tmp/pylint-xblock.log
 ERR=`grep -E "^[C|R|W|E]:" /tmp/pylint-xblock.log | wc -l`
--- a/script/max_pylint_violations
+++ b/script/max_pylint_violations
@@ -1,10 +1,10 @@
 #!/bin/bash
-DEFAULT_MAX=11
+DEFAULT_MAX=$15
 
 pylint xblock | tee /tmp/pylint-xblock.log
 ERR=`grep -E "^[C|R|W|E]:" /tmp/pylint-xblock.log | wc -l`
 MAX=${1-$DEFAULT_MAX}
-if [ $ERR -ge $MAX ]; then
+if [ $ERR -gt $MAX ]; then
     echo "too many pylint violations: $ERR (max is $MAX)"
     exit 1
 else
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -6,7 +6,7 @@ and used by all runtimes.
 
 """
 from __future__ import unicode_literals
-from builtins import bytes
+from builtins import bytes  # pylint: disable=redefined-builtin
 import future.utils
 
 import inspect
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -287,7 +287,7 @@ class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixi
         If all of the aside's data is empty or a default value, then the aside shouldn't
         be serialized as XML at all.
         """
-        return any([field.is_set_on(self) for field in self.fields.itervalues()])
+        return any(field.is_set_on(self) for field in self.fields.values())
 
 
 # Maintain backwards compatibility
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -7,14 +7,14 @@ and used by all runtimes.
 """
 from __future__ import unicode_literals
 from builtins import bytes  # pylint: disable=redefined-builtin
-import future.utils
 
 import inspect
-import pkg_resources
 import warnings
 import os
 from collections import defaultdict
 
+import future.utils
+import pkg_resources
 from xblock.exceptions import DisallowedFileError
 from xblock.fields import String, List, Scope
 from xblock.internal import class_lazy
@@ -30,6 +30,9 @@ from xblock.mixins import (
 )
 from xblock.plugin import Plugin
 from xblock.validation import Validation
+
+# This is imported just to re-expose exceptions that were moved into that file
+import xblock.exceptions
 
 # exposing XML_NAMESPACES as a member of core, in order to avoid importing mixins where
 # XML_NAMESPACES are needed (e.g. runtime.py).
@@ -295,10 +298,6 @@ class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixi
         be serialized as XML at all.
         """
         return any(field.is_set_on(self) for field in future.utils.itervalues(self.fields))
-
-
-# Maintain backwards compatibility
-import xblock.exceptions
 
 
 class KeyValueMultiSaveError(xblock.exceptions.KeyValueMultiSaveError):
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -5,6 +5,9 @@ This code is in the Runtime layer, because it is authored once by edX
 and used by all runtimes.
 
 """
+from builtins import bytes
+import future.utils
+
 import inspect
 import pkg_resources
 import warnings
@@ -89,6 +92,9 @@ class SharedBlockBase(Plugin):
         matched against a whitelist regex to ensure that you do not serve an
         unauthorized resource.
         """
+
+        if isinstance(uri, bytes):
+            uri = uri.decode('utf-8')
 
         # If no resources_dir is set, then this XBlock cannot serve local resources.
         if cls.resources_dir is None:
@@ -287,7 +293,7 @@ class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixi
         If all of the aside's data is empty or a default value, then the aside shouldn't
         be serialized as XML at all.
         """
-        return any(field.is_set_on(self) for field in self.fields.values())
+        return any(field.is_set_on(self) for field in future.utils.itervalues(self.fields))
 
 
 # Maintain backwards compatibility
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -5,6 +5,7 @@ This code is in the Runtime layer, because it is authored once by edX
 and used by all runtimes.
 
 """
+from __future__ import unicode_literals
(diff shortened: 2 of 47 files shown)


cpennington:
@jcdyer, @nedbat, @maxrothman: This is up to date, now.

xblock/test/test_core.py
```diff
@@ -3,16 +3,18 @@
 Tests the fundamentals of XBlocks including - but not limited to -
 metaclassing, field access, caching, serialization, and bulk saves.
 """
+from __future__ import unicode_literals
+from builtins import next, str  # pylint: disable=redefined-builtin
```
jcdyer: Do you know why reimporting `next` is required?  That one should just work in python 2, right?

xblock/test/toy_runtime.py
```diff
@@ -83,7 +84,7 @@ def set_many(self, update_dict):
         `update_dict`: A dictionary of keys: values.
         This method sets the value of each key to the specified new value.
         """
-        for key, value in update_dict.items():
+        for key, value in list(update_dict.items()):
```
jcdyer: This should probably use `future.utils.iteritems()`.

script/max_pylint_violations
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
-DEFAULT_MAX=11
+DEFAULT_MAX=$15
```
jcdyer: Should this be `$1`?

xblock/test/test_core.py
cpennington: The python 2 version won't use `__next__`, I think.

cpennington:
After this exercise, it's not clear to me whether it's better to use `from __futue__ import unicode_literals`, or to go through and explicitly mark every string. The latter is a significant increase in the amount of work (even though I prefer it, from a correctness point of view).

jcdyer:
Checking to see if edx-platform's tests pass with this branch: https://github.com/edx/edx-platform/pull/13263

xblock/test/test_core.py
jcdyer: Huh.  That's what the docs say it's good for, but that hypothetical code that only defines `__next__` (and not `next`) will fail as soon as python 2 tries to use it in a loop.  Seems minimally useful, but no big deal.

xblock/reference/__init__.py
```diff
@@ -3,3 +3,4 @@
 
 See README.md for details.
 """
+from __future__ import unicode_literals
```
jcdyer: On an empty `__init__`?  Wouldn't it be easier just to put a `u` in front of the docstring?

xblock/run_script.py
```diff
@@ -13,6 +14,6 @@ def run_script(pycode):
 
     # execute it.
     globs = {}
-    exec pycode in globs, globs  # pylint: disable=W0122
+    exec(pycode, globs, globs)  # pylint: disable=W0122
```
jcdyer: Won't this break python2?  I'm pretty sure I had to do a `six.exec_()` in my port to work around exec incompatibilities.

xblock/runtime.py
```diff
@@ -1267,17 +1268,42 @@ def __getattr__(self, name):
...
         """
         format = self.STRFTIME_FORMATS.get(format + "_FORMAT", format)
-        if isinstance(format, unicode):
-            format = format.encode("utf8")
-        return dtime.strftime(format).decode("utf8")
+        if future.utils.PY2 and isinstance(format, str):
+            format = format.encode("utf-8")
+        timestring = dtime.strftime(format)
+        if future.utils.PY2:
+            timestring = timestring.decode('utf-8')
```
jcdyer: I feel like all these custom chunks that branch on `future.utils.PY2` should be pulled out into an edx-future library.  I think we'll need them in a number of places.

xblock/test/test_core.py
```diff
@@ -821,10 +828,10 @@ class HasParent(XBlock):
 
 def test_json_handler_basic():
     test_self = Mock()
-    test_data = {"foo": "bar", "baz": "quux"}
+    test_data = {u"foo": u"bar", u"baz": u"quux"}
     test_data_json = json.dumps(test_data)
-    test_suffix = "suff"
-    test_request = Mock(method="POST", body=test_data_json)
+    test_suffix = u"suff"
+    test_request = Mock(method=u"POST", body=test_data_json)
```
jcdyer: Nit: unicode_literals are set on this file.

xblock/core.py
```diff
@@ -287,11 +297,7 @@ def needs_serialization(self):
         If all of the aside's data is empty or a default value, then the aside shouldn't
         be serialized as XML at all.
         """
-        return any([field.is_set_on(self) for field in self.fields.itervalues()])
-
-
-# Maintain backwards compatibility
-import xblock.exceptions
+        return any(field.is_set_on(self) for field in future.utils.itervalues(self.fields))
```
nedbat: I thought one of the advantages of future was that it didn't put stuff like "future.utils.itervalues" in your code? I can see it would be difficult to avoid this automatically.

xblock/django/request.py
```diff
@@ -76,9 +80,15 @@ def querydict_to_multidict(query_dict, wrap=None):
 
     """
     wrap = wrap or (lambda val: val)
+
+    if future.utils.PY2:
+        item_lists = query_dict.iterlists()
+    else:
+        item_lists = query_dict.lists()
```
nedbat: For expected-short lists, wouldn't it be better to just use the common method name that makes a list in Py2?

xblock/field_data.py
```diff
@@ -87,7 +87,7 @@ def set_many(self, block, update_dict):
         :param update_dict: A map of field names to their new values
         :type update_dict: dict
         """
-        for key, value in update_dict.items():
+        for key, value in future.utils.iteritems(update_dict):
```
nedbat: Here we can certainly keep `.items()`, no?

xblock/fields.py
```diff
@@ -465,7 +473,7 @@ def _calculate_unique_id(self, xblock):
         for the field in its given scope.
         """
         key = scope_key(self, xblock)
-        return hashlib.sha1(key).hexdigest()
+        return hashlib.sha1(key.encode('utf-8')).hexdigest()
```
nedbat: How does this work in Python 2? The old code must have had `key` as bytes, is it now Unicode?

xblock/fields.py
```diff
@@ -1013,22 +1042,22 @@ def scope_key(instance, xblock):
...
         pass
     elif instance.scope.user == UserScope.ONE:
-        scope_key_dict['user'] = unicode(xblock.scope_ids.user_id)
+        scope_key_dict['user'] = str(xblock.scope_ids.user_id)
     else:
         raise NotImplementedError()
 
     if instance.scope.block == BlockScope.TYPE:
-        scope_key_dict['block'] = unicode(xblock.scope_ids.block_type)
+        scope_key_dict['block'] = str(xblock.scope_ids.block_type)
```
nedbat: So `str` here is Unicode in Python 2? That seems really confusing.

xblock/runtime.py
```diff
@@ -1267,17 +1268,42 @@ def __getattr__(self, name):
...
+        return timestring
+
+    @property
+    def ugettext(self):
+        """
+        Dispatch to the appropriate gettext method
+        """
+        # pylint: disable=no-member
+        if future.utils.PY3:
```
nedbat: Since we are starting fresh here, the conditionals should all be on PY2, not PY3, for that day in the far future when Python 4.x is out, and behaves like Python 3 (which we have been promised it will).

xblock/core.py
cpennington: No way to avoid it without monkey-patching (which I don't want to do).

xblock/django/request.py
cpennington: Yes, that would probably be better.

xblock/field_data.py
cpennington: Yeah.

xblock/fields.py
cpennington: Just remember, the goal is Py3 code that runs the same on Py2, so `str` is the Py3 `str`.

xblock/runtime.py
cpennington: Yup, good catch.

xblock/fields.py
cpennington: Hrm. I'll have to take a look at that. We may need some additional enforcement in the `scope_key` function.

jcdyer: I believe what's happening is that python2 will autoconvert unicode to bytes (using the ascii codec), and this code relied on that.  Of course that would have broken on non-ascii characters, but scope_key seems to be pretty careful to restrict its output set.

xblock/fields.py
nedbat: I understand the goal. Another goal is, don't drive the developers crazy as they try to work in this polyglot world. It might be better to have an obvious indicator that things are 2/3-complicated.  If I dive into a repo and see "str", how do I determine what it means? I have to keep my eye on the top of the file to know what paradigm I'm using: pure 2, pure 3, 2/3 with futurize, "2/3 with futurize but forgot to import str", etc.  I think six has an advantage that its polyglot nature is clearer with a smaller search radius.

xblock/fields.py
nedbat: @jcdyer Yes, that seems to be it.

xblock/fields.py
jcdyer: ^ This is why the *other* xblock conversion PR exists (#349): to surface the differences between a `six`-based approach, and a `future`-based approach, so we can better decide which one we want to advocate to the company.  

I think one of the issues here is that while futurize will do the initial conversions for you, it becomes easy to forget things during later development.  If I'm working in a file that doesn't currently use `str()`, and I add it, have I added the future import to the top of the file?  

Six requires more developer awareness and understanding of the differences up front, but is more explicit about where polyglot code lives.

PR #356, opened 2016-11-07
[WIP] Introduce UI blocks

andy-armstrong:
This is a proof-of-concept of a new variant of XBlocks currently called UIBlocks.

You can read about the rationale on the wiki here:

https://openedx.atlassian.net/wiki/display/TNL/UIBlocks+for+Course+Outline

### DIFF
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -105,6 +105,20 @@ class SharedBlockBase(Plugin):
 
         return pkg_resources.resource_stream(cls.__module__, os.path.join(cls.resources_dir, uri))
 
+    @property
+    def supports_save(self):
+        """
+        True if this block supports being saved.
+        """
+        return False
+
+    @property
+    def supports_asides(self):
+        """
+        True if this block can be shown with asides.
+        """
+        return False
+
 
 # -- Base Block
 class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeServicesMixin, HandlersMixin,
@@ -188,9 +202,12 @@ class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeS
         # Provide backwards compatibility for external access through _field_data
         super(XBlock, self).__init__(runtime=runtime, scope_ids=scope_ids, field_data=field_data, *args, **kwargs)
 
-    def render(self, view, context=None):
-        """Render `view` with this block's runtime and the supplied `context`"""
-        return self.runtime.render(self, view, context)
+    @property
+    def supports_asides(self):
+        """
+        True if this block can be shown with asides.
+        """
+        return True
 
     def validate(self):
         """
@@ -221,7 +238,7 @@ class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeS
 
 class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixin, HandlersMixin, SharedBlockBase):
     """
-    This mixin allows Xblock-like class to declare that it provides aside functionality.
+    This class allows XBlock-like classes to declare that they provides aside functionality.
     """
 
     entry_point = "xblock_asides.v1"
@@ -288,6 +305,44 @@ class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixi
         be serialized as XML at all.
         """
         return any([field.is_set_on(self) for field in self.fields.itervalues()])
+
+
+# -- UI-only Block
+class UIBlock(RuntimeServicesMixin, HandlersMixin, ViewsMixin, SharedBlockBase):
+    """Base class for UI Blocks.
+
+    This class differs from XBlocks in that they are intended only to
+    provide pluggable user interfaces. They explicitly have no state
+    and cannot be persisted.
+
+    Derive from this class to create a new kind of UIBlock.  There are no
+    required methods, but you will probably need at least one view.
+
+    Don't provide the ``__init__`` method when deriving from this class.
+
+    """
+    entry_point = 'ui_block.v1'
+
+    def __init__(self, block_type, runtime, *args, **kwargs):
+        """
+        Construct a new UIBlock.
+
+        This class should only be instantiated by runtimes.
+
+        Arguments:
+
+            runtime (:class:`.Runtime`): Use it to access the environment.
+                It is available in XBlock code as ``self.runtime``.
+
+            field_data (:class:`.FieldData`): Interface used by the XBlock
+                fields to access their data from wherever it is persisted.
+                Deprecated.
+
+            scope_ids (:class:`.ScopeIds`): Identifiers needed to resolve
+                scopes.
+        """
+        super(UIBlock, self).__init__(runtime=runtime, *args, **kwargs)
+        self.block_type = block_type
 
 
 # Maintain backwards compatibility
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -105,20 +105,6 @@ class SharedBlockBase(Plugin):
 
         return pkg_resources.resource_stream(cls.__module__, os.path.join(cls.resources_dir, uri))
 
-    @property
-    def supports_save(self):
-        """
-        True if this block supports being saved.
-        """
-        return False
-
-    @property
-    def supports_asides(self):
-        """
-        True if this block can be shown with asides.
-        """
-        return False
-
 
 # -- Base Block
 class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeServicesMixin, HandlersMixin,
@@ -201,13 +187,6 @@ class XBlock(XmlSerializationMixin, HierarchyMixin, ScopedStorageMixin, RuntimeS
 
         # Provide backwards compatibility for external access through _field_data
         super(XBlock, self).__init__(runtime=runtime, scope_ids=scope_ids, field_data=field_data, *args, **kwargs)
-
-    @property
-    def supports_asides(self):
-        """
-        True if this block can be shown with asides.
-        """
-        return True
 
     def validate(self):
         """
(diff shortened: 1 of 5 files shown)


robrap:
@andy-armstrong: On this PR, is your intention to have this approved as a successful proof-of-concept, or to get this merged?

Also, @andy-armstrong @nasthagiri @cpennington @ormsbee, do we think the concept of a UIBlock requires an OEP?  I can imagine arguments on both sides.

andy-armstrong:
@robrap Good question. For now, I'm just looking for feedback on the approach and the implementation. If folks are agreed that this is the right thing to do, I'll work on productizing this ticket by adding tests, fixing quality issues etc.

xblock/core.py
```diff
@@ -105,6 +105,20 @@ def open_local_resource(cls, uri):
 
         return pkg_resources.resource_stream(cls.__module__, os.path.join(cls.resources_dir, uri))
 
+    @property
+    def supports_save(self):
```
robrap: Would it be enough to just add a no-op `save()` call here and drop `supports_save()`?

xblock/core.py
```diff
@@ -221,7 +238,7 @@ def add_xml_to_node(self, node):
 
 class XBlockAside(XmlSerializationMixin, ScopedStorageMixin, RuntimeServicesMixin, HandlersMixin, SharedBlockBase):
     """
-    This mixin allows Xblock-like class to declare that it provides aside functionality.
+    This class allows XBlock-like classes to declare that they provides aside functionality.
```
robrap: Typo: "...they provides..."

robrap:
I think the Runtime needs clear documentation and/or naming conventions (e.g. `block`, `xblock`, `uiblock`) to help understand what methods can handle all blocks (e.g. `render()`) and what methods work with `uiblock` only.  

It may clarify the proposal to add a comment to those methods that are `uiblock` specific, and those that we are proposing would support both.

xblock/mixins.py
```diff
@@ -242,6 +249,13 @@ def _field_data(self, field_data):
         warnings.warn("Setting _field_data is deprecated", FieldDataDeprecationWarning, stacklevel=2)
         self._deprecated_per_instance_field_data = field_data
 
+    @property
+    def supports_save(self):
+        """
+        Returns true if this block supports being saved.
+        """
+        return True
```
ormsbee: If I'm understanding this correctly, we're putting this here so that the runtime code can generically call render on XBlocks that want to save data after a view is executed, and UIBlocks that do not support that functionality. But if that's the case, maybe we could make the runtime not knowing about saving at all, but more generically execute post-rendering hooks? XBlock would then define post-rendering to do a `save()`, while UIBlocks would just do a no-op?

xblock/runtime.py
```diff
@@ -650,6 +650,19 @@ def get_block(self, usage_id, for_parent=None):
         block = self.construct_xblock(block_type, keys, for_parent=for_parent)
         return block
 
+    def load_ui_block_type(self, block_type):
+        """
+        Returns a subclass of :class:`.UIBlock` that corresponds to the specified `block_type`.
+        """
+        return UIBlock.load_class(block_type, select=self.select)
```
ormsbee: Could `load_block_type()` be made more generic to handle both?

ormsbee:
To piggy-back on @robrap's comment about naming conventions... I think that if we're proposing that we may have more block families (combinations of mixins) beyond XBlocks, XBlockAsides, and UIBlocks -- then we have to try to figure out how much of that knowledge can be pushed out of the core Runtime and into the block families themselves.

The danger in one direction is that the Runtime ends up with a bunch of similar methods repeated for each new family of things, making it more cumbersome to experiment with new types. The danger in the other is that we pile on into a mass of undebuggable *args/**kwargs soup.

Ideally, I feel like discovery (loading of classes), identity (how do I refer to the one instance of this thing for lookup and URL purposes), and rendering should be handled by the same code path.

xblock/mixins.py
robrap: I was thinking about this too, but it is a bit more complicated because `block.save()` is called 3 times in the runtime, and `aside.save()` is also called twice.  

One of the other `block.save()` calls is in `handle`, which I'm guessing applies to UIBlocks, but I'm not certain.  This code would need to change if we want the runtime to be less concerned with the need to save.  

The final call to `block.save()` is in `_usage_id_from_node` inside the runtime.  If we really want more independence between the mixin and runtime, this might require an even deeper change.

Additionally, the `aside.save()` calls could be ignored because they are in aside-specific runtime methods, or we could again try to match models of separating concerns out of the runtime.

Maybe I shouldn't have, but due to this I decided to punt and request that save just be added as a no-op.

xblock/core.py
nasthagiri: If the presence of this Mixin in the class hierarchy indicates Aside support, then the `supports_asides` method is superfluous.

xblock/core.py
```diff
@@ -290,6 +307,44 @@ def needs_serialization(self):
         return any([field.is_set_on(self) for field in self.fields.itervalues()])
 
 
+# -- UI-only Block
+class UIBlock(RuntimeServicesMixin, HandlersMixin, ViewsMixin, SharedBlockBase):
+    """Base class for UI Blocks.
+
+    This class differs from XBlocks in that they are intended only to
+    provide pluggable user interfaces. They explicitly have no state
+    and cannot be persisted.
```
nasthagiri: .. for now.

Eventually, I do think we'll want to add support for persistence and state for UIBlocks as well.

nasthagiri:
As a follow-on of @ormsbee and @robrap comments:

When I first proposed (a form of) UIBlocks [in August](https://docs.google.com/a/edx.org/drawings/d/1NLb_rCzT3M4XOV3ODYHu1hKaQ1SJ8YHsf4lUNXZwkag/edit?usp=sharing), I was thinking they would be part of a class hierarchy of ViewBlocks.

Since then, however, I've been thinking more along the lines of allowing developers to mix-n-match Mixins to construct their own Block variations.  By doing so, it may result in a more powerful toolset.  However, we'll need to keep the possible explosion of permutations manageable and well-named.

If we follow this design pattern, then the runtime wouldn't need to worry about all specific types of blocks.  Rather, it can use duck-typing (by the presence/absence of Mixins) to determine supported behavior.

And then along those lines, as others have mentioned, it would be great if we don't have a proliferation of factory methods within the runtime itself.  Rather, the runtime can have a single factory method for all block types, and any block-type specific creation can be encapsulated within a known class method on the block-type's class.

cpennington:
We'd likely need some introspection gymnastics to make sure that the various families of blocks got the right arguments for their constructors, but it does seem like an avenue worth exploring.

xblock/core.py
andy-armstrong: IMO we don't want to have persistence for UI Blocks. I think of them as the view part of a typical model-view-controller paradigm, so if they need persisted data that would live in one of the models that it has access to. Maybe we are saying the same thing, in that the UI Block would have access to persisted data through runtime services.

So to your original comment here: I'll update the docstring to say that UI blocks have no state or persistence, but that they have access to the data they need through runtime services.

xblock/runtime.py
```diff
@@ -973,13 +987,14 @@ def render_asides(self, block, view_name, frag, context):
...
         """
         aside_frag_fns = []
-        for aside in self.get_asides(block):
-            aside_view_fn = aside.aside_view_declaration(view_name)
-            if aside_view_fn is not None:
-                aside_frag_fns.append((aside, aside_view_fn))
-        if aside_frag_fns:
-            # layout (overideable by other runtimes)
-            return self.layout_asides(block, context, frag, view_name, aside_frag_fns)
+        if block.supports_asides:
```
andy-armstrong: The issue was that asides assumes that it has access to a ``usage_id`` in the block it is rendering on behalf of, and that was deeply embedded into the logic such that it seemed simpler to just not try to call it. IMO it makes sense that XBlock asides only render for XBlocks. I'll see if there's a more Pythonic way of doing this, as was discussed with my other ``supports_save`` property.

xblock/core.py
andy-armstrong: I considered the no-op ``save`` but that seems ugly to me. I intended to make this a class attribute and didn't think through what ``@property`` meant, so I'll switch it and see how that seems.

xblock/mixins.py
andy-armstrong: I really like @ormsbee's proposal so that the XBlock class itself provides the desired save. I'll try that out and see if that is sufficient.

xblock/core.py
cpennington: Also, I would point out that, in fact, scope-storage *is* backed by a runtime service (called 'field-data').

robrap:
Additional in-progress documentation available here: https://openedx.atlassian.net/wiki/display/FEDX/UI+Blocks

andy-armstrong:
@robrap @ormsbee @nasthagiri @cpennington I worked on a new approach that introduces a UI Block runtime, rather than extending the existing runtime to work with it. I believe that this idea was suggested to @robrap by @cpennington and it makes a lot of the code cleaner:
 - there can be a shared UI runtime to render in any context, while XBlock runtimes are tied to a course
 - the XBlock runtime doesn't get polluted with UI Block-only methods
 - I could remove all the special handling of asides and saving for UI blocks as the new runtime doesn't include that logic

I'm going to write up a UI Block OEP today and point to this PR as the proposed technical implementation. No rush, but I'd love your feedback on the new approach.

cpennington:
I don't think we need a new runtime class. I'd rather see the current runtime decomposed into its constituent pieces (xml, hierarchy, view rendering), or something like that. Unless there's truly no common logic...

cpennington:
When I suggested a new runtime, I really was just thinking about a new runtime implementation in the LMS, not a new base-class in XBlock.

andy-armstrong:
Closing in favor of https://github.com/andy-armstrong/django-component-views/pull/1

PR #357, opened 2016-12-01
Add XMLString field type which validates input.

jcdyer:
### [TNL-5245](https://openedx.atlassian.net/browse/TNL-5245)

Add XMLString field type which validates that input is real XML.

This will be used to ensure inbound capa data is valid xml.  See reference ticket for rationale.

### Reviewers

- [x] @sanfordstudent 
- [x] @doctoryes

FYI @efischer19 @nasthagiri


xblock/__init__.py
```diff
@@ -22,4 +22,4 @@ def __init__(self, *args, **kwargs):
 
 xblock.fields.XBlockMixin = XBlockMixin
 
-__version__ = "0.4.7"
+__version__ = "0.5.0"
```
doctoryes: Nit: It'd be nice to pull the `setup.py` version and this one from a common location, particularly considering that this version was out-of-date.

xblock/fields.py
```diff
@@ -832,17 +834,24 @@ class String(JSONField):
...
         Leave all other characters.
         """
-        if isinstance(value, unicode):
-            new_value = u''.join(ch for ch in value if unicodedata.category(ch)[0] != u'C' or ch in (u'\n', u'\r', u'\t'))
-        elif isinstance(value, str):
-            new_value = ''.join(ch for ch in value if ord(ch) >= 32 or ch in ('\n', '\r', '\t'))
+        if isinstance(value, six.text_type):
+            new_value = u''.join(ch for ch in value if self._valid_unichar(ch))
+        elif isinstance(value, six.binary_type):
+            new_value = ''.join(ch for ch in value if self._valid_bytechar(ch))
```
doctoryes: nice refactor!

xblock/test/test_fields.py
```diff
@@ -244,6 +241,34 @@ def test_control_characters_filtered(self):
...
+        xml_string = self.FIELD_TO_TEST()
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, 'text')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<xml')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<xml attr=3/>')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<xml attr="3/>')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<open>')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<open>with text')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<xml/>trailing text')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<open>text</close>')
+        self.assertRaises(etree.XMLSyntaxError, xml_string.from_json, '<invalid_utf8 char="\x9e"/>')
```
doctoryes: Both of these tests would be better off using DDT'd lists of test data.

xblock/__init__.py
jcdyer: Done.

xblock/test/test_fields.py
jcdyer: Done.

jcdyer:
max_pylint_violations script had non-obvious behavior: matching the max number of errors was a failure.  Should be exceeding max number is an error.

jcdyer:
Ready for another round of review.

jcdyer:
@cpennington Here's the PR to create the xml type.

xblock/VERSION.txt
```diff
@@ -0,0 +1 @@
+0.5.0
```
cpennington: This isn't backwards incompatible, right? No need to go to 0.5.0 over 0.4.13, right?

jcdyer:
@cpennington This has been updated to validate when saving, rather than when loading, which should reduce the impact of the checks.Take a look.

jcdyer:
@sanfordstudent @doctoryes This now does validation when serializing instead of deserializing, and only does it if enforce_type is set to True.  That way open edx installations can temporarily turn off the feature to fix bad data.

sanfordstudent:
Still looks good to me.

PR #359, opened 2017-01-24
Separate Fragment class out into new web-fragments package

andy-armstrong:


xblock/fragment.py
```diff
@@ -1,269 +1,9 @@
...
-        elif resource.mimetype == "text/html":
-            assert resource.kind == "text"
-            return resource.data
+from web_fragments.fragment import Fragment
 
-        else:
-            raise Exception("Never heard of mimetype %r" % resource.mimetype)
+__all__ = [
+    'Fragment',
+]
```
nedbat: There's no need for `__all__` here. It only affects "from xblock.fragment import *", and without it, the same names would be imported anyway.

setup.py
```diff
@@ -30,6 +30,7 @@
         'pyyaml',
         'six',
         'webob',
+        'web-fragments',
```
nedbat: We'll have to be careful not to use this version of XBlock until web-fragments is on PyPI.

xblock/fragment.py
andy-armstrong: So I can just have a file with an import and nothing else? Does that seem like the Pythonic way to achieve what I want?

setup.py
andy-armstrong: Agreed. I definitely won't merge this until web fragments 0.1.0 has been published.

xblock/fragment.py
nedbat: A comment would help explain what you are doing, and you might need a pylint pragma.  But it will work.

cpennington: Yeah, there are actually other spots in XBlock where I did the same thing while moving classes around internally.

setup.py
andy-armstrong: FYI, the PR for web fragments is: https://github.com/andy-armstrong/web-fragments/pull/1

pmitros:
Thank you. This looks nice. I might consider adding a second (non-squashed) commit which removes fragment.py, and changes imports globally. This commit makes it abundantly clear what has changed. However, as a static snapshot of code, extra layers of indirection make code hard to follow.

setup.py
andy-armstrong: Web fragments is now on PyPI: https://pypi.python.org/pypi/web-fragments/0.1.0

andy-armstrong:
@pmitros The problem with removing fragment.py is that we have XBlocks that have been developed which use the ``Fragment`` class. Once this new version goes live in the LMS, it will break all of those blocks. I think we're stuck with leaving the delegating code for the time being.

xblock/fragment.py
andy-armstrong: I've removed the ``__all__`` and added a comment.

andy-armstrong:
@nedbat This PR is ready for your re-review when you have a chance. @cpennington can you be the second reviewer? Thanks!

pmitros:
@andy-armstrong Makes sense. 

I would still consider changing internal imports, and adding a deprecation message if fragment.py is imported from xblock rather than the appropriate repo.

andy-armstrong:
@pmitros thanks, that makes sense. BTW, what is the Pythonic way to add a deprecation message? I couldn't find any examples in our code base.

cpennington:
@andy-armstrong: https://github.com/edx/XBlock/blob/6fe0a6d4de2fc629fc7adfc6f30ca96ab7f912a0/xblock/__init__.py#L18

pmitros:
https://docs.python.org/2/library/exceptions.html#exceptions.DeprecationWarning

Or, if one is lazy, one can just print:

     print("Warning: You are importing from xblock.fragment instead of fragments.fragment. This"
     print("this is deprecated. Please update your code. See [URL of PR, OEP, or similar] for 
     print("details.")

If one is fancy, I'd also print where I was being imported from. 

@nedbat

andy-armstrong:
Thanks @cpennington and @pmitros. I was searching for deprecated, so missed DeprecationWarning.

andy-armstrong:
@pmitros @cpennington @nedbat - I've taken @pmitros's advice to deprecate the Fragment class and to change the code to use the version from web-fragments. I had to also add a small unit test just to keep the code coverage up. 

Please re-review when you have a moment. If possible I'd like to merge tomorrow before our sprint closes out. Thanks.

pmitros:
Looks good to me. :+1: 

As a pedantic footnote: If I had done this, I would not have subclassed Fragment. I would have written this as:

      import warnings
      from fragments.Fragment import Fragment

      warnings.warn(....)

This has two advantages:

- If any code relies on `type`, `isinstance()`, etc., it doesn't break
- We only print the warning once, on application startup, rather than a whole stream of warnings. It's visible earlier, and doesn't pollute logs. 

That said, it's interim code, until XBlocks update, so I'm not sure it's worth changing.

cpennington:
I agree w/ @pmitros about not subclassing, but for different reasons. I would have put the warning at the module level so that it would trigger if anyone imported `xblock.fragments`, no matter what they did with it after that.

`warnings.warn` does a fairly good job at only warning a limited number of times, so either way shouldn't flood the logs too badly (I think it will warn exactly once for each code path that hits the warning, so you get all of the unique stack traces).

andy-armstrong:
@pmitros @cpennington that's ironic, as I switched it to a class based upon the other deprecated example! I liked it as I thought it made the delegating attributes somewhat cleaner, rather than adding them to the web fragments class itself. If in the future the web-fragments class changes more substantially, it also allows us to add back missing methods etc. 

Let me know if you'd like me to switch it back, or if it is okay as it stands. Thanks!

PR #361, opened 2017-02-16
This is a dependent repo, and no longer needs this clause

nedbat:


nedbat:
@cpennington this clause was here to support tagging this repo during Open edX releases. We no longer tag dependent repos.

PR #362, opened 2017-03-01
ScorableXBlockMixin

jcdyer:
# [TNL-5511](https://openedx.atlassian.net/browse/TNL-5511)

## Reviewers

- [x] @sanfordstudent 
- [x] @nasthagiri


xblock/scorable.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def rescore(self, only_if_higher):
+        """
+        Calculate a new score and save it to the block.  If only_if_higher is
+        True and the score didn't improve, keep the existing score.
+
+        Returns True if the score was changed.
+        Returns False if the score was not changed.
+        Raises a TypeError if the block cannot be scored.
+        Raises a ValueError if the user has not yet completed the problem.
```
jcdyer: Is it useful for the caller to know if this changed a score or not?

xblock/scorable.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        return {
+            'usage_key': six.text_type(self.location)
+        }
+
+    def allows_rescore(self):
+        """
+        Boolean value: Can this problem be rescored.
+        """
+        return True
```
nasthagiri: Should this check whether the `get_score`,  `set_score` and `calculate_score` methods are implemented?

xblock/scorable.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def calculate_score(self):
+        """
+        Calculate a new score based on the state of the problem.
+        This method should should not modify the state of the
+        XBlock.
+
+        Returns:
+            Score(earned=float, total=float)
+        """
```
nasthagiri: raise `NotImplementedError` here?

xblock/scorable.py
nasthagiri: Or, rather, why not just have the `NotImplementedError` be raised when `rescore` is called?

xblock/scorable.py
```diff
@@ -0,0 +1,143 @@
...
+
+        Returns True if the score was changed.
+        Returns False if the score was not changed.
+        Raises a TypeError if the block cannot be scored.
+        Raises a ValueError if the user has not yet completed the problem.
+        """
+
+        _ = self.runtime.service(self, 'i18n').ugettext
+        if not self.allows_rescore():
+            self._publish_scoring_event('rescore_failure', {'failure': 'unsupported'})
```
jcdyer: Agreed.  I was slow on the uptake between event publishing and logging, because Capa publishes events for everything, but I'm not sure we need to do that.

xblock/scorable.py
jcdyer: NotImplementedError is specially used by pylint to indicate abstract methods that haven't been overridden.  It is intended to signal missing code, not runtime logic violations.  Additionally, if we try to use it for runtime checking, we'll need to silence pylint everywhere, which is less than ideal.  

We also can't just let the error bubble up, because there are reasons to disallow rescore on problem types that have those methods implemented.  The Capa XModule has the ability for the lcp to block rescoring for certain problems.  I haven't looked to deeply into why, but I assume it's used for survey problems at least.

nasthagiri: The [python documentation](https://docs.python.org/3/library/exceptions.html#NotImplementedError) states that `NotImplementedError` can be raised to indicate unfinished work (in addition to requiring subclasses implement abstract methods).  So if keeping the code simple and elegant requires silencing pylint, I'm all for it.  That is, if pylint gets in the way, I'd rather remove the rule or quiet it.

However, I had forgotten about how CAPA problems can be configured to allow/not-allow rescoring.  So if the purpose of this method is NOT for whether the functionality is implemented, but rather, to allow subclasses to providing permission/configuration options, then what you have here is fine.  Can you update the docstring, however, to explain this use case?

xblock/scorable.py
```diff
@@ -0,0 +1,158 @@
...
+            Calculate a new score and save it to the block.
+            `only_if_higher
+
+        ScorableXBlockMixin.allows_rescore(self) -> bool
+            Returns True if the block can be rescored.
+
+    Subclasses must define the following:
+
+        ScorableXBlockMixin.get_score(self)
+            Return the currently saved score.  Must not calculate a new score.
```
sanfordstudent: These docstrings should specify weighted vs. raw. While theoretically an XBlock shouldn't know its own weight, the drag-and-drop v2 XBlock knows its own weight and stores its grade as a weighted value. While any block could choose to implement get_, set_, and calculate_score with either raw or weighted scores (as long as they were consistent), for our logging and debugging purposes I think that it is useful to know what data we are looking at and have it be consistent.

nasthagiri: +1

nasthagiri: @sanfordstudent I remember reviewing the drag-n-drop code in a previous PR earlier where this came up.  Are you sure they store their score as a weighted value?  That sounds like a significant issue - since a problem weight change normally doesn't require a re-storing of all scores.

sanfordstudent: @nasthagiri I may be following the code incorrectly. It's here: https://github.com/edx-solutions/xblock-drag-and-drop-v2/blob/master/drag_and_drop_v2/drag_and_drop_v2.py#L907

`self.grade` is stored as a weighted value. When a grade is published, the code uses `self.grade` and `self.weight` for the earned and possible values.

xblock/scorable.py
jcdyer: Given that the only_if_higher calculation is pushed out to the grading subsystem, I think we can just not return from this method, and let the caller follow up with a call to get_score() if they want to know the new value of the score.

xblock/core.py
```diff
@@ -5,11 +5,12 @@
 and used by all runtimes.
 
 """
+from collections import defaultdict
```
jcdyer: They aren't new imports, just reordering of the existing ones.  I'll revert if it causes merge problems.

xblock/scorable.py
jcdyer: That's a good idea.

xblock/scorable.py
```diff
@@ -0,0 +1,143 @@
...
+
+        _ = self.runtime.service(self, 'i18n').ugettext
+        if not self.allows_rescore():
+            self._publish_scoring_event('rescore_failure', {'failure': 'unsupported'})
+            raise TypeError(_('Problem does not support rescoring'))
+
+        original_score = self.get_score()
+        if original_score is None:
+            self._publish_scoring_event('rescore_failure', {'failure': 'unanswered'})
+            raise ValueError(_('Problem must be answered before it can be rescored.'))
```
jcdyer: Yes, but it raises an `xmodule.exceptions.NotFoundError`, which we obviously don't have access to here.

xblock/scorable.py
jcdyer: Great investigation.  I've specified raw scores all around.

xblock/scorable.py
```diff
@@ -0,0 +1,152 @@
...
+        self.runtime.publish('grade', grade_dict)
+
+    def _publish_scoring_event(self, name, data):
+        """
+        Publish an event related to scoring.
+        """
+        logged = {}  # Ensure we don't mutate our arguments
+        logged.update(data)
+        logged.update(self._base_event_info())
+        self.runtime.publish(name, logged)
```
sanfordstudent: I think you might need to pass `self` as the first parameter to `self.runtime.publish` here and elsewhere. See https://github.com/edx/XBlock/blob/master/xblock/runtime.py#L503

jcdyer: You are correct.

jcdyer:
Ready for another round of review.

xblock/scorable.py
```diff
@@ -0,0 +1,140 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from collections import namedtuple
+import logging
+
+
+log = logging.getLogger(__name__)
+
+
+Score = namedtuple('Score', ['earned', 'total'])
```
nasthagiri: Per @sanfordstudent's suggestion, can we be explicit and consistent with these names?  So: `raw_earned, raw_possible` instead of `earned, total`?  This is the terminology I introduced in https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/graders.py#L47.

xblock/scorable.py
```diff
@@ -0,0 +1,140 @@
...
+
+class ScorableXBlockMixin(object):
+    """
+    Mixin to handle functionality related to scoring.
+
+    Provides the following public methods:
+
+        ScorableXBlockMixin.rescore(self, only_if_higher)
+            Calculate a new score and save it to the block.
+            `only_if_higher
```
nasthagiri: Incomplete docstring?

nasthagiri: Instead of needing to maintain the docstrings for these methods in 2 different places, don't include them in the docstring of the `class`.  Just include them in each method's docstring below.

xblock/scorable.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Calculate a new raw score and save it to the block.  If only_if_higher
+        is True and the score didn't improve, keep the existing score.
+
+        Raises a TypeError if the block cannot be scored.
+        Raises a ValueError if the user has not yet completed the problem.
+        """
+
+        _ = self.runtime.service(self, 'i18n').ugettext
+        errorlog_prefix = _('Rescore failure:')
```
nasthagiri: The `log` statements in this method seem superfluous and will cause duplication of logs.  It's sufficient to simply raise the error with the correct message and let higher callers log.  See the caller's implementation in https://github.com/edx/edx-platform/blob/master/lms/djangoapps/instructor_task/tasks_helper.py#L479

xblock/scorable.py
```diff
@@ -0,0 +1,140 @@
...
+
+        original_score = self.get_score()
+        if original_score is None:
+            error_msg = _('Cannot rescore unanswered problem: {}').format(self.location)
+            log.error('{} {}'.format(errorlog_prefix, error_msg))
+            raise ValueError(error_msg)
+
+        try:
+            new_score = self.calculate_score()
+        except Exception as exc:  # pylint: disable=broad-except
```
nasthagiri: See my earlier comment.  No reason to handle this exception at this layer.  I would advocate letting the exception go through and let it be logged in the instructor-task database.

xblock/scorable.py
```diff
@@ -0,0 +1,140 @@
...
+            error_msg = _('A calculation error occurred when rescoring {}: {}').format(self.location, exc)
+            log.warning('{} {}'.format(errorlog_prefix, error_msg))
+            return
+
+        self.set_score(new_score)
+        self._publish_grade(only_if_higher)
+        if only_if_higher:
+            only_if_higher_msg = 'Rescore only if score improved.'
+        else:
+            only_if_higher_msg = 'Rescore unconditionally'
```
nasthagiri: These statements will not be internationalized - the statement below will, but not these phrases.

xblock/scorable.py
jcdyer: Ok.  Updated.  The only log statement left is the rescore result.  No other place captures both the previous and new scores.

xblock/scorable.py
sanfordstudent: +1

jcdyer:
Ready again.

jcdyer:
Bumped xblock version so we can pull it into edx-platform.

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class ScorableXBlockMixin(object):
+    """
+    Mixin to handle functionality related to scoring.
+
+    Provides the following public methods:
+
+        ScorableXBlockMixin.allows_rescore(self) -> bool
+            Returns True if the block can be rescored.
```
nasthagiri: Reminder to remove the duplicated doc-strings of the methods.

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+
+        Raises a TypeError if the block cannot be scored.
+        Raises a ValueError if the user has not yet completed the problem.
+        """
+
+        _ = self.runtime.service(self, 'i18n').ugettext
+        resultlog_prefix = _('Rescore result:')
+
+        if not self.allows_rescore():
+            error_msg = _('Problem does not support rescoring: {}').format(self.location)
```
nasthagiri: Nit: a temporary variable for a constant error message doesn't seem necessary.

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+            raise TypeError(error_msg)
+
+        original_score = self.get_score()
+        if original_score is None:
+            error_msg = _('Cannot rescore unanswered problem: {}').format(self.location)
+            raise ValueError(error_msg)
+
+        new_score = self.calculate_score()
+
+        self.set_score(new_score)
```
nasthagiri: The setting of the score is done within the grades layer (after comparing the score if `only_if_higher`): https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/signals/handlers.py#L118

So this is redundant and incorrect if the grade is not equal or higher.  I suggest removing this line.

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+        if only_if_higher:
+            only_if_higher_msg = _('Rescore only if score improved.')
+        else:
+            only_if_higher_msg = _('Rescore unconditionally.')
+        result_msg = _('Original score: {}. New score: {}. {}').format(
+            original_score,
+            new_score,
+            only_if_higher_msg
+        )
+        log.info('{} {}'.format(resultlog_prefix, result_msg))
```
nasthagiri: Given that the comparison and setting of the scores is done in the grades layer, I suggest moving this log statement there - as that layer owns whether or not the rescore actually happens.
On the other hand, I don't see the need for this log statement.  The instructor task already emits an event for this: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/instructor_task/tasks_helper.py#L580

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+        rescoring.
+        """
+        return True
+
+    def get_score(self):
+        """
+        Return a raw score already persisted on the XBlock.
+
+        Returns:
+            Score(raw_earned=float, raw_possible=float)_
```
nasthagiri: Nit: Extra `_` character at the end.

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        raise NotImplementedError
+
+    def set_score(self, score):
+        """
+
```
nasthagiri: Nit: extra blank line.

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+        Persist a score to the XBlock.
+
+        The score is a named tuple with a raw_earned attribute and a
+        max_possible attribute, reflecting the raw earned score and the maximum
```
nasthagiri: s/max_possible/raw_possible

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+
+        Returns:
+            None
+        """
+        raise NotImplementedError
+
+    def calculate_score(self):
+        """
+        Calculate a new raw score based on the state of the problem.
+        This method should should not modify the state of the
```
nasthagiri: s/should should/should

xblock/scorable.py
```diff
@@ -0,0 +1,131 @@
...
+        Returns:
+            Score(raw_earned=float, raw_possible=float)
+        """
+        raise NotImplementedError
+
+    def _publish_grade(self, only_if_higher=None):
+        """
+        Publish a grade to the runtime.
+        """
+        score = self.get_score()
```
nasthagiri: Do not expect `get_score` to be a fast operation.  As it is right now, it is getting called 3-times (!).  Once here, once in the caller of this method, once again in the grades layer.

xblock/scorable.py
jcdyer: You're talking about setting the *grade*.  Setting the score is about persisting the calculated result to the xblock, (not about updating the graded result), which, if you look at how capa does it, is done irrespective of the only_if_higher calculation.

xblock/scorable.py
```diff
@@ -0,0 +1,122 @@
...
+    def allows_rescore(self):
+        """
+        Boolean value: Can this problem be rescored?
+
+        Subtypes may wish to override this if they need conditional support for
+        rescoring.
+        """
+        return True
+
+    def has_score(self):
```
sanfordstudent: This might collide with a current XBlock attribute that simply says whether the block _can_ have a score, not whether it does. It isn't well-documented in this repo but it's referred to here. https://github.com/edx/XBlock/blob/5b6af7a8cde09bd103ccd91919a858b6f088e28a/xblock/runtime.py#L507 

Drag and drop already sets a `has_score` attribute, and it's always `True`.

jcdyer:
Ready for another review with new simplified implementation.

jcdyer:
Hmm.  I don't like the "is_committed".  It's only clear in the context of scoring, but not in the general context of being an xblock method.  I'm going to change it again.  What do you think about "has_committed_answer"?

sanfordstudent:
Do we want to add `has_score = True` to the mixin by default? After all, if it's scorable, it had better have a score.

jcdyer:
@sanfordstudent Done.

xblock/scorable.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Subclasses must define the following:
+
+        ScorableXBlockMixin.has_submitted_answer(self)
+        ScorableXBlockMixin.get_score(self)
+        ScorableXBlockMixin.set_score(self, score)
+        ScorableXBlockMixin.calculate_score(self, score)
+    """
+
+    has_score = True
```
nasthagiri: Nit: Can you add a docstring to explain the semantic of this class variable and why we need it?

xblock/scorable.py
```diff
@@ -0,0 +1,114 @@
...
+        Boolean value: Can this problem be rescored?
+
+        Subtypes may wish to override this if they need conditional support for
+        rescoring.
+        """
+        return True
+
+    def has_submitted_answer(self):
+        """
+        Returns true if the problem has been answered before.
```
nasthagiri: s/`answered before`/`answered by the runtime user before`

PR #366, opened 2017-09-14
Increase the tox test matrix to include different Django versions

bmedx:


xblock/test/django/test_request.py
```diff
@@ -24,6 +24,7 @@
 
 from webob import Response
 
+from xblock import django  # pylint: disable=unused-import
```
doctoryes: Just curious why this import was needed?

bmedx: Something, probably a new version of coverage, coverage-diff, or a change on codecov caused `xblock.django.__init__.py` to no longer be imported, causing a tiny coverage hit that was blocking merge.

PR #368, opened 2017-10-18
CompletionXBlockMixin + tests

e-kolpakov:
**Description:** This PR adds `CompletableXBlockMixin` mixin to allow XBlocks use [Completion API](https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/162247762/Completion+API)

**JIRA tickets**: N/A (Part of [Completion API](https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/162247762/Completion+API) work)

**Discussions**: See Completion API document above.

**Dependencies**: **Dependencies:** https://github.com/edx/edx-platform/pull/16234

**Merge deadline**: Required for the feature with deadline at Nov 27th, 2017 - so no hard deadline, but would be nice to have it merged by Oct 20th, 2017

**Author concerns:** I'm using [hypothesis](https://hypothesis.readthedocs.io/en/latest/quickstart.html) for property-based testing, but the same can be achieved (likely with more code and covering less cases) with ddt. I find property-based testing a very powerful tool, but it introduces a test dependency (pretty lightweight though) and I haven't explicitly discussed that with anyone. I don't expect to find any significant opposition, but would be nice to vet this decision.

The only non-negligible drawback is that it introduces quite different test style, so to keep "homogenity" it might be desired to stick with ddt. However, as you can see in this PR prop-based testing also supports data-driven test (called table-driven property-checks in that camp), so the two styles can peacefully coexist.

**Testing Instructions:**

1. Run tests (should be handled by Travis)

**Reviewers**
- [x] @jcdyer
- [ ] @robrap


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2FXBlock&number=368"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

jcdyer:
I'll leave the question about hypothesis to the folks at edX.  Personally, I like the way it looks.  Curious about how many tests it generates and how it affects test suite running time.

Will take a closer look at this soon.

xblock/VERSION.txt
```diff
@@ -1 +1 @@
-1.0.1
+1.1.1
```
jcdyer: Nit: Should this be 1.1.0?

xblock/test/test_mixins.py
```diff
@@ -473,3 +478,58 @@ def test_datetime_serialization(self, value, expected_attributes):
...
+    def _make_block(self, runtime=None):
+        """ Creates a test block """
+        runtime = runtime if runtime else mock.Mock(spec=Runtime)
+        scope_ids = ScopeIds("user_id", "test_buddy", "def_id", "usage_id")
+        return self.TestBuddyXBlock(runtime=runtime, field_data=DictFieldData({}), scope_ids=scope_ids)
+
+    def test_has_custom_completion_property(self):
+        block = self._make_block()
+        self.assertTrue(block.has_custom_completion)
+        self.assertTrue(getattr(block, 'has_custom_completion', False))
```
jcdyer: What's the benefit of getting the block both through direct attribute access and through getattr?

xblock/test/test_mixins.py
```diff
@@ -473,3 +478,58 @@ def test_datetime_serialization(self, value, expected_attributes):
...
+        block = self._make_block()
+        self.assertEqual(block.completion_method, 'completable')
+        self.assertEqual(getattr(block, 'completion_method', ""), 'completable')
+
+    @given(strategies.floats(min_value=0.0, max_value=1.0))
+    @example(1.0)
+    @example(0.0)
+    def test_emit_completion_emits_event(self, valid_completion_ratio):
+        """
+        Given valid completion ratio
```
jcdyer: Grammar: Given *a* valid completion ratio

xblock/test/test_mixins.py
```diff
@@ -473,3 +478,58 @@ def test_datetime_serialization(self, value, expected_attributes):
...
+        block.emit_completion(valid_completion_ratio)
+
+        runtime_mock.publish.assert_called_once_with(block, "completion", {"completion": valid_completion_ratio})
+
+    @given(strategies.floats().filter(lambda x: math.isnan(x) or x < 0.0 or x > 1.0))
+    @example(float('+inf'))
+    @example(float('-inf'))
+    def test_emit_completion_raises_assertion_error_if_invalid(self, invalid_completion_ratio):
+        """
+        Given invalid completion ratio
```
jcdyer: Grammar: Given *an* invalid completion ratio

xblock/test/test_mixins.py
```diff
@@ -473,3 +478,58 @@ def test_datetime_serialization(self, value, expected_attributes):
...
+            * Less than 0.0
+            * Greater than 1.0
+            * Positive or negative infinity
+            * NaN
```
jcdyer: Maybe include `None`, since that's a common value to accidentally pass in.

xblock/mixins.py
```diff
@@ -610,3 +611,19 @@ def has_support(self, view, functionality):
             True or False
         """
         return hasattr(view, "_supports") and functionality in view._supports  # pylint: disable=protected-access
+
+
+class CompletableXBlockMixin(object):
```
jcdyer: This and the emit_completion method should have docstrings.  The class docstring should include a brief note regarding why it should be used. (For blocks that need special control over when they are considered complete).  The method docstring should explain that emitting a value of 1.0 indicates that a block has been completed, and that emitting a value of 0.0 on a previously completed block indicates that it is no longer considered complete.

xblock/mixins.py
```diff
@@ -610,3 +611,19 @@ def has_support(self, view, functionality):
...
+    has_custom_completion = True
+    completion_method = 'completable'
+
+    def emit_completion(self, completion_ratio):
```
jcdyer: I think this method should verify that has_custom_completion is `True` and completion_method is `'completable'`.  The value is set here, but I can imagine a scenario where a subclass wants to fall back to default completion calculation.  If that subclass sets `has_custom_completion = False`, it will fall back to one of the default completion implementations, but also emit its own completion events, which will lead to unexpected behavior.

A test that emit_completion exits silently in those cases would be useful.

xblock/mixins.py
e-kolpakov: @jcdyer I don't think it should exit silently - this mixin doesn't have anything useful besides this method - so falling back to default should be performed by not mixin in this class.

xblock/mixins.py
e-kolpakov: @jcdyer I don't think it should exit silently - this mixin doesn't have anything useful besides this method - so falling back to default should be performed by not mixing in this class.

jcdyer: If the mixin is not a direct superclass, butfurther up the inheritance chain, that won't be an option.  Even though it makes the mixin effectively a noop, that is still the behavior of least surprise and last frustration. The only other options are to cause an error, which will just irritate some poor developer, or to emit the completion anyway, which is exactly the opposite behavior you'd expect when setting has_custom_completion to False.

e-kolpakov: @jcdyer Let me clarify something - we're discussing behavior of custom completion method _when it is called_ on non-custom-compketion block. To me, it is an outright programming error and should be indicated to programmer assoon as possible.

So to me NoCustomCompletion.emit_completion()  is programming error. I realize this mixing can be inherited from some other block, but I can't see a scenario where emit_completion would be rightfully called on non-custom-compketion completion block. If you could describe such scenario I'll gladly revoke this objection. 

In any case, if you really strongly believe silently doing nothing is preferred over raising an error and taking into account you'll be doing majority of work on this feature I'll just stick with your preferred way. However I'm not (yet) convinced it is an optimal long-term solution.

xblock/mixins.py
```diff
@@ -610,3 +611,43 @@ def has_support(self, view, functionality):
...
         return hasattr(view, "_supports") and functionality in view._supports  # pylint: disable=protected-access
+
+
+class CompletableXBlockMixin(object):
+    """
+    This mixin sets attributes and provides helper method to integrate XBlock with Completion API.
+    """
+
+    has_custom_completion = True
+    completion_method = 'completable'
```
jcdyer: I mentioned on the ticket that I got a request from edx to change completion method to an enum or a set of constants.  Would you mind adding that to this PR?  Needed values:

* completable
* aggregator
* excluded

xblock/mixins.py
```diff
@@ -610,3 +611,43 @@ def has_support(self, view, functionality):
...
+
+    def emit_completion(self, completion_ratio):
+        """
+        Emits completion event through Completion API.
+
+        Unlike grading API, calling this method allows completion to go down - i.e. emitting a value of 0.0 on
+        a previously completed block indicates that it is no longer considered complete.
+
+        Arguments:
+            completion_ratio (float): Ratio of completion within [0.0; 1.0] range (inclusive). 0.0 means the block
```
robrap: We agreed to user the term "percent" instead of "ratio".  In the model, this is just called "completion".  I'm option to "completion", "percent_complete", and "completion_percent", but I'd like to be consistent where and if this appears again (if it hasn't already), so maybe @jcdyer can weigh in.  

Note: For consistency, I'd change the doc string to read: "Percent of completion within [0.0; 1.0] range..."

xblock/mixins.py
```diff
@@ -610,3 +611,43 @@ def has_support(self, view, functionality):
...
+        """
+        if not self.has_custom_completion or self.completion_method != 'completable':
+            raise AttributeError(
+                "Using `emit_completion` requires `has_custom_completion == True` (was {}) "
+                "and `completion_method == 'completable'` (was {})".format(
+                    self.has_custom_completion, self.completion_method
+                )
+            )
+
+        if completion_ratio is None or not 0.0 <= completion_ratio <= 1.0:
```
robrap: Can you drop `completion_ratio is None or` ?

xblock/test/test_mixins.py
```diff
@@ -473,3 +478,109 @@ def test_datetime_serialization(self, value, expected_attributes):
...
+        """
+        runtime_mock = mock.Mock(spec=Runtime)
+        illegal_completion_method_block = self._make_block(runtime_mock, self.TestIllegalCompletionMethodAttrXBlock)
+        with self.assertRaises(AttributeError):
+            illegal_completion_method_block.emit_completion(any_completion)
+
+    @given(strategies.floats(min_value=0.0, max_value=1.0))
+    @example(1.0)
+    @example(0.0)
+    def test_emit_completion_emits_event(self, valid_completion_ratio):
```
robrap: Reminder: when we decide on the proper terms, search the entire PR for "ratio" and replace with the "percent"-like terms.  Thanks.

xblock/mixins.py
e-kolpakov: @robrap Please forgive me being pedantic, but "percent" is in range [0.0; 100.0], not [0.0; 1.0]. So just renaming the parameter but keeping allowed range would be extremely misleading. Please consider either:

1. Keeping this code intact and renaming other occurences to use mathematically correct term.
2. Change both name and allowed range (if I'm not mistaking it also involves changes down the stream).

Of course I can just rename it to "percent", but it would be a misleading name.

xblock/mixins.py
e-kolpakov: @jcdyer I think I've lost track of this request - will do tomorrow.

xblock/mixins.py
e-kolpakov: @robrap It'll cause python 3 tests to fail with TypeError or something like that. Basically python 3 is more picky when it comes to comparing None to anything else.

xblock/mixins.py
jcdyer: @e-kolpakov I'm totally with you on the distinction, and had originally named it "ratio" on my design doc, but we determined that it's more important to be consistent with edx-platform's general terminology, which is to call floats constrained to [0.0, 1.0] "percent".

robrap: We've discussed this at length in other conversations, and I think we should continue as previously decided.  

A brief summary:
1. Ratio isn't precise either, because a ratio could be any range, and it is the fact that it is a percent that requires the range of 0 to 1.
2. I don't think it is "extremely" misleading.  There are places in our code as well as other code bases where a percent is stored or passed around as 0 to 1, even though it will finally become 0-100 when it is displayed with a percent sign.

Worst case scenario, someone coding with this for the first time doesn't read the docs, thinks it is 0 to 100, and quickly finds out they are wrong when 1% is the max percent they are able to reach.

xblock/mixins.py
robrap: Got it.  Thank you.

xblock/mixins.py
e-kolpakov: @robrap Is american math or something? :) Ok, I'll do that, but since I have "minority report" I'll add a link to this discussion.

Regarding design doc - I probably haven't reached the part where you guys discuss "ratio" vs. "percent".

Ok, final two thing before I wrap up this discussion:

1. Percent can be any range also (130% ROI, -15% market price drop, etc.).
2. Ratio is the accurate term here - ratio of 1.0 means "things you're comparing are exactly equal", which in our case corresponds to "you've got maximum score possible".
3. The worst case scenario you described contributes to bad developer experience. Basically having ratios called percents anywhere edx-platform causes:
3.1. The need to translate name into actual meaning each time someone touches the code operating on grades/progress/etc. - i.e. taxes productivity a little bit.
3.2. And taxes productivity heavier when the scenario you described happens - which is actually not limited to first use, but might also be caused by forgetting the fact, rushing for whatever reason (i.e. hotfix) and maybe some more.
3.3. Long and pointless discussions like this one :)

So, as a member of community I flag this "percent" vs. "ratio" problem as something that reduces my productivity. It's just a tiny bit, but probably you guys (core developers and architects) might consider this as part of technical debt and address when/if there's opportunity to do so.

And sure, I understand effort & risk vs. outcome is somewhere in range "not a chance", but still.

e-kolpakov:
Rebased from 5808436

xblock/test/test_core.py
```diff
@@ -1039,6 +1039,13 @@ class TestBlock(XBlock):
         """An empty XBlock for testing"""
         pass
 
+    # FIXME: This test is fragile - fails in python27 if a warning it is expecting is emitted before it is executed
```
e-kolpakov: @robrap FYI.

bdero: The warnings module has different default filter behavior for different warning categories across different versions of python. Most likely DeprecationWarning is set to ["once"](https://docs.python.org/2/library/warnings.html#the-warnings-filter) in this case.

I think this can be worked around by calling `warnings.resetwarnings()` or `warnings.simplefilter("always")` in the test setup.

jcdyer: TIL.  Thanks @bdero

e-kolpakov: @jcdyer @bdero No guys, it's [not that simple](https://github.com/edx/XBlock/blob/master/xblock/test/tools.py#L118).

bdero: @e-kolpakov Have you tried sliding a `warnings.resetwarnings()` in there before the catch_warnings line? It's totally possible there's a subtle behavior difference between different versions of python (or a bug in the specific release being used), so hopefully playing around with the warning options will yield some good information on the problem.

e-kolpakov: @bdero Yes, I've tried (and double checked just now) spamming with `resetwarnings` and `filter("always")` with different combinations and at different locations - no help.

Unfortunately, I don't have capacity right now to further investigate this issue, so I'd suggest we just wrap it up as is.

@robrap It's not a flaky test - it is a stable test that's easy to break as it depends on the order of tests executed. In short, it only passes now because `test_core` is run before any other test that could trigger the deprecation warning. If I understand right, nose runs tests alphabetically, so test_core is run very early - only asides and now completion is run earlier. Asides code does not have anything to do with creating XBlocks (where the deprecation warning is set), and completion also doesn't use the deprecated feature - but only after I figured out what's wrong with it - most of the other tests are still using it.

e-kolpakov:
@robrap Addressed/commented on your notes, please take another look.

Github UI might be confusing a bit - when one responds to a comment from "Files" tab and clicks "Start my review" / "Add Comment" it displays the response both in the original thread (with an option to respond to it) and as part of review (without). So, answer to your question about test flakiness is a bit above ([here](https://github.com/edx/XBlock/pull/368#discussion_r147029117)) :)

xblock/test/test_core.py
jmbowman: @e-kolpakov @robrap I suspect that sometime reasonably soon we'll want to switch this repository to use pytest instead of nose, and make a concerted effort to fix any dependencies between the tests so they can be run in random order.  If the tests are stable until the order changes, I think it's probably ok to leave it as is until we tackle that.  pytest has some dedicated functionality for testing warnings which may prove useful when we attempt that.

robrap: Thanks @e-kolpakov.  You may want to add an additional note regarding @jmbowman's response in the comment if you think it would be helpful to someone coming across this.  Or, you could leave a link to this discussion regarding how and when it might get fixed.

e-kolpakov:
@robrap Thanks - merged. I'm sorry about how it looks on master now - despite squash merging (and fixing those small typos in commit messages) github merged it as is for some reason. Do you think it's worth the trouble squashing them on master (+force push and all that unpleasant things)?

robrap:
I don't think it is worth the trouble, as long as master now has what you imagined it should have.  I usually squash and let the tests run on the PR, so I'm not sure how often github doesn't do what you want.

e-kolpakov:
@robrap Actually it doesn't - I wanted it to be a single commit, but it just added all of them without squashing.

robrap:
Right.  I just imagined that at least sometimes "squash merge" actually squashes before merging.  In any case, it seems like use-at-your-own-risk functionality.

e-kolpakov:
@robrap Normally it squashes, but this time something went wrong. Anyway, I'd prefer to squash them on master to have relatively clean history, so unless you (or someone) have objections I'll do that tomorrow morning (it's almost midnight in my timezone)

robrap:
@e-kolpakov: I don't think you should mess with master until I hear back on our practices.  I think the branch is protected and you may not be able to force push to master anyway.  It seems our best practices would typically have "squash and merge" disabled, so I'm looking into that.  See "Protected Branches" and "Merge Button" sections of this doc: https://openedx.atlassian.net/wiki/spaces/IT/pages/82576458/GitHub+Repository+Admin+Guidelines

robrap:
FYI: @e-kolpakov: Our policy is to not mess with master and we'll just leave this as-is.  In the future, you may want to rely on squashing locally to ensure you'll get what you want.  Thanks again for the changes.

PR #372, opened 2017-11-20
PLAT-1413 Remove obsolete TEMPLATE_* settings

jmbowman:
Removed obsolete (and unused) TEMPLATE_* settings.  I also noticed that we haven't been running quality checks in Travis for a while due to a tox environment name mismatch; I broke those out into a separate tox environment and fixed the minor new issues it reported.  The violation threshold did increase by one, due to a new `FIXME` comment in `xblock/test/test_core.py` which will take some investigation (and possibly a switch to pytest) to resolve.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,6 +6,12 @@ rednose=1
 # Uncomment the following line to open pdb when a test fails
 #pdb=1
 
-[pep8]
-ignore=E501,E402
+[pycodestyle]
+# error codes: https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
+# E402: module level import not at top of file
+#   We catch this with pylint, don't need 2 reports of the same issue
+# E501: line too long
+# E722: do not use bare except, specify exception instead
+#   We catch this with pylint, don't need 2 reports of the same issue
+ignore=E501,E402,E722
 exclude=doc/*,.tox/
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,6 @@ deps =
     -rrequirements.txt
 commands =
     coverage run -m nose {posargs}
-    py27-django18: make quality
 whitelist_externals =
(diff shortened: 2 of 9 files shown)


setup.cfg
```diff
@@ -6,6 +6,12 @@ rednose=1
...
-[pep8]
-ignore=E501,E402
+[pycodestyle]
+# error codes: https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
+# E402: module level import not at top of file
+#   We catch this with pylint, don't need 2 reports of the same issue
+# E501: line too long
+# E722: do not use bare except, specify exception instead
+#   We catch this with pylint, don't need 2 reports of the same issue
+ignore=E501,E402,E722
```
doctoryes: So pycodestyle is a drop-in replacement for PEP-8? Is it our new best practice?

jmbowman: The package was renamed at Guido's request: https://github.com/pycqa/pycodestyle

> This package used to be called pep8 but was renamed to pycodestyle to reduce confusion. Further discussion can be found in the issue where Guido requested this change, or in the lightning talk at PyCon 2016 by @IanLee1521: slides video.

tox.ini
```diff
@@ -31,3 +30,10 @@ deps=
     -rdoc/requirements.txt
 commands =
     make html
+
+[testenv:quality]
+deps =
+    Django>=1.11,<2.0
```
pwnage101: what's the motivation for selecting this django version for running quality tests?

jmbowman: I just had to pick one for imports to succeed, and this is the one we plan to be on shortly.  It shouldn't really matter in this context.

jmbowman:
The tiny coverage drop is because the only change to relevant code was deletion of covered lines.  Merging.

PR #376, opened 2018-06-08
 Don't translate lazily-translated default string value.

doctoryes:
Not happy with this impl yet. 
I'm looking for your input into how to properly allow lazily translated default values for XBlock String fields.

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -1,0 +1,21 @@
+import os
+import sys
+os.environ['DJANGO_SETTINGS_MODULE'] = 'xblock.test.settings'
+
+from django.conf import settings
+settings.configure(
+    DEBUG=True,
+    ALLOWED_HOSTS='*'
+)
+
+import django
+django.setup()
+
+# Now we instantiate a test runner...
+from django.test.utils import get_runner
+TestRunner = get_runner(settings)
+
+# And then we run tests and return the results.
+test_runner = TestRunner(verbosity=2, interactive=True)
+failures = test_runner.run_tests(['xblock.test.django.test_field_translation'])
+sys.exit(failures)
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{27,35}-django{18,110,111}
+envlist = py{27,36}-django{18,110,111}
 
 [testenv]
 passenv = CI TRAVIS TRAVIS_*
@@ -12,9 +12,10 @@ deps =
     django18: Django>=1.8,<1.9
     django110: Django>=1.10,<1.11
     django111: Django>=1.11,<2.0
-    -r requirements/test.txt
+    -rrequirements/test.txt
 commands =
     coverage run -m nose {posargs}
+    coverage run -m runner.py
 whitelist_externals =
      make
 
--- a/xblock/fields.py
+++ b/xblock/fields.py
@@ -434,6 +434,12 @@ class Field(Nameable):
         baseline = xblock._dirty_fields[self]
         return baseline is EXPLICITLY_SET or xblock._field_data_cache[self.name] != baseline
 
+    def _is_lazy(self, value):
+        """
+        Detect if a value is being evaluated lazily by Django.
+        """
+        return 'django.utils.functional.__proxy__' in str(type(value))
+
     def _check_or_enforce_type(self, value):
         """
         Depending on whether enforce_type is enabled call self.enforce_type and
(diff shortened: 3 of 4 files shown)


runner.py
```diff
@@ -0,0 +1,21 @@
...
+django.setup()
+
+# Now we instantiate a test runner...
+from django.test.utils import get_runner
+TestRunner = get_runner(settings)
+
+# And then we run tests and return the results.
+test_runner = TestRunner(verbosity=2, interactive=True)
+failures = test_runner.run_tests(['xblock.test.django.test_field_translation'])
+sys.exit(failures)
```
jmbowman: It looks like this didn't quite work as expected in Travis.  I'll go ahead and start a pytest/pytest-django conversion PR, may be easier than trying to get this debugged.

tox.ini
```diff
@@ -4,17 +4,18 @@
...
+envlist = py{27,36}-django{18,110,111}
 
 [testenv]
 passenv = CI TRAVIS TRAVIS_*
 deps =
     django18: Django>=1.8,<1.9
     django110: Django>=1.10,<1.11
     django111: Django>=1.11,<2.0
-    -r requirements/test.txt
+    -rrequirements/test.txt
```
jmbowman: The space is allowed as of tox 3.0.0, and better matches typical pip command line usage.

xblock/fields.py
```diff
@@ -454,14 +460,16 @@ def _check_or_enforce_type(self, value):
...
-            try:
-                equal = value == new_value
-            except TypeError:
-                equal = False
-            if not equal:
-                message = "The value {!r} would be enforced to {!r}".format(
-                    value, new_value)
-                warnings.warn(message, ModifyingEnforceTypeWarning, stacklevel=3)
+            # Refuse to evaluate lazy evaluations during this type enforcement.
+            if not self._is_lazy(value) and not self._is_lazy(new_value):
```
jmbowman: Would it make more sense to check instead if one of the values is lazy, and if so coerce both to unicode text before doing the comparison?  That would just move the condition check inside the try block, leaving the rest of the method unchanged from before.  Is there a problem with the default value being evaluated before the i18n machinery is set up?

runner.py
doctoryes: I agree. That's the direction we want to go anyway. As you saw, this technique requires making a separate test runner and running tests in two different ways. Using pytest alone along with pytest-django is better.

tox.ini
doctoryes: Ah - I wondered about that!

xblock/fields.py
doctoryes: Exactly - this code avoids the evaluation of the lazily translated values before Django's i18n functionality is ready. Coercing or comparing them when one or both is lazily translated always ended up evaluating and causing this error:
```
======================================================================
1) ERROR: xblock.test.django.test_field_translation.test_db_model_keys
----------------------------------------------------------------------
    .tox/py27-django111/lib/python2.7/site-packages/nose/case.py line 197 in runTest
      self.test(*self.arg)
    xblock/test/django/test_field_translation.py line 49 in test_db_model_keys
      assert_equals(tester.str_field, None)
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/functional.py line 148 in __eq__
      return self.__cast() == other
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/functional.py line 131 in __cast
      return self.__text_cast()
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/functional.py line 119 in __text_cast
      return func(*self.__args, **self.__kw)
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/translation/__init__.py line 89 in ugettext
      return _trans.ugettext(message)
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/translation/trans_real.py line 345 in ugettext
      return do_translate(message, 'ugettext')
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/translation/trans_real.py line 321 in do_translate
      _default = _default or translation(settings.LANGUAGE_CODE)
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/translation/trans_real.py line 228 in translation
      _translations[language] = DjangoTranslation(language)
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/translation/trans_real.py line 129 in __init__
      self._add_installed_apps_translations()
    .tox/py27-django111/lib/python2.7/site-packages/django/utils/translation/trans_real.py line 171 in _add_installed_apps_translations
      "The translation infrastructure cannot be initialized before the "
   AppRegistryNotReady: The translation infrastructure cannot be initialized before the apps registry is ready. Check that you don't make non-lazy gettext calls at import time.
```

doctoryes:
@jmbowman Feel free to close this PR whenever you'd like - in favor of the pytest-django approach.

doctoryes:
Closed in favor of https://github.com/edx/XBlock/pull/381

PR #377, opened 2018-06-11
TE-2604 Switch from nose to pytest

jmbowman:
Switch the test runner to fix some issues testing the use of Django translated strings and to be consistent with our other repositories.  Notable changes required:

* Applied `__test__ = False` to some classes which are named like test classes but actually aren't.  This avoids some warnings about classes which weren't collected because they had `__init__` methods.
* Configured pytest to always output all warnings.  This fixed a flaky test, but also means we now get 369 warnings at the end of the run because we hadn't been displaying DeprecationWarnings at all before.  We should fix those, but that seems like it should be done in separate tickets.
* For test classes which don't inherit from `unittest.TestCase`, pytest uses `setup_method` instead of `setUp`.  Renamed and fixed pylint warnings accordingly.


jmbowman:
@doctoryes Could you take a look at that `pylint --py3k` warning and see if you think it has any merit?  I can't reproduce it locally, and the code doesn't seem to suffer from the indicated problem.  I'm inclined to just instruct pylint to ignore it if we can't find a genuine issue.

doctoryes:
@jmbowman What's the path forward for codecov here? Should we make a new branch, cherry-pick, and new PR? Or can we unstick codecov?

jmbowman:
I'll try closing and reopening the PR.  If that fails, I'll try renaming to a new branch and open a new PR for that.  I can override the required status for the codecov check if necessary.

doctoryes:
🎉

PR #380, opened 2018-06-15
Fix codecov data reporting

jmbowman:
The workaround for pytest finding tests both in the root directory and the installed packages left the collected coverage data file in the wrong place.  Move it where expected and see if this fixes codecov reports.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,8 +22,10 @@ deps =
 changedir = {envsitepackagesdir}
 commands =
     pytest {posargs:xblock}
+    mv .coverage {toxinidir}/.coverage
 whitelist_externals =
     make
+    mv
 
 [testenv:docs]
 basepython =


jmbowman:
@doctoryes So this seems to have mostly fixed the codecov issue, but 2 of the 16 Travis job results went AWOL on their way to codecov...or rather, some sort of S3 permission error is preventing codecov from accessing them:

https://codecov.io/gh/edx/XBlock/commit/01ee2de8bffb9f5b1037e5bcec53b54f7922a809
https://codecov.io/gh/edx/XBlock/commit/01ee2de8bffb9f5b1037e5bcec53b54f7922a809/build

I can tell codecov to go ahead and send the report so far to GitHub without those data points, but this doesn't exactly make me confident in requiring codecov for merges again...

doctoryes:
Does codecov have a process for alerting them to issues? Seems like this problem should be a bug report for them.

jmbowman:
There's a support email address, but I wasn't sure whether to report things like this as I find them or send a more comprehensive report of issues once we've collected data from https://openedx.atlassian.net/browse/TE-2603 .

doctoryes:
Yes - comprehensive seems better.

PR #426, opened 2019-07-22
 Make Mixologist able to handle class path strings.

feanil:
Previously you could only pass in class objects to the Mixologist.  This
meant that the mixin classes had to be imported somewhere else.  This
allows us to defer importing of mixin classes to when the Mixologist is
instantiated.

The reason for this need is that in edx-platform we currently import the
mixins as a part of the settings loading which casues issues in python
3. See BOM-66 for more details.


xblock/runtime.py
```diff
@@ -1211,10 +1212,22 @@ class Mixologist(object):
...
         :type mixins: `iterable` of `class`
         """
-        self._mixins = tuple(mixins)
+
+        mixin_classes = []
+        for mixin in mixins:
+            if isinstance(mixin, six.text_type):
+                module, cls = mixin.rsplit('.', 1)
+                imported_module = importlib.import_module(module)
+                # Should I catch the attribute error and raise a clearer error on failure?
```
nedbat: I'd say it would be nice to catch errors (including from the rsplit) so that mistakes in the class name would look like "Couldn't import class foobar.mistake.baz: AttributeError: 'module' object has no attribute 'baz'".

nedbat:
Don't forget to update CHANGELOG.rst

xblock/runtime.py
```diff
@@ -1211,10 +1212,22 @@ class Mixologist(object):
     """
     def __init__(self, mixins):
```
ormsbee: Maybe we should make a new param for the paths instead of overloading `mixins` to be a list of either kind?

xblock/runtime.py
```diff
@@ -1211,10 +1212,22 @@ class Mixologist(object):
...
         """
-        :param mixins: Classes to mixin
+        :param mixins: Classes to mixin or names of classes to mixin.
         :type mixins: `iterable` of `class`
         """
-        self._mixins = tuple(mixins)
+
+        mixin_classes = []
+        for mixin in mixins:
+            if isinstance(mixin, six.text_type):
```
ormsbee: I think a lot of folks will get bit by putting `['my.class.name']` in Python 2.7 without `unicode_literals` and then blowing this up because it's a byte string and not unicode.

xblock/runtime.py
feanil: I considered that but went with this since the order of the mixins matters.

xblock/runtime.py
feanil: This is only used in one place in edx-platform for now.  But you're suggesting I use [six.string_types](https://six.readthedocs.io/#six.string_types) instead?

xblock/runtime.py
ormsbee: Ah, I see. Works for me then.

xblock/runtime.py
ormsbee: Yes, but point taken that the failure scenario isn't as likely as I originally thought.

ormsbee:
So yeah, no asks beyond what @nedbat has already flagged in his review.

FYI @cpennington

feanil:
Ready for re-review then.

xblock/runtime.py
```diff
@@ -1211,10 +1212,26 @@ class Mixologist(object):
...
+                module, cls = mixin.rsplit('.', 1)
+                imported_module = importlib.import_module(module)
+                try:
+                    mixin_class = getattr(imported_module, cls)
+                except AttributeError as e:
+                    msg = "Couldn't import class {}: {}".format(mixin, str(e))
```
nedbat: This try-clause should include lines 1222 and 1223 also, since they can also indicate problems with the class name.
Also: no need for `str(e)`, `e` would give the same result, since `{}` will str it for you.

If you expand the scope of the except, then we don't know what exception to catch, and you'd want to include the class in the message:
```
except Exception as e:
    msg = "Couldn't import class {!r}: {}: {}".format(mixin, e.__class__.__name__, e)
    raise ImportError(msg)
```

nedbat:
Still need to update CHANGELOG

cpennington:
No additional comments over and above what's already been posted.

xblock/runtime.py
feanil: sounds good, I'll make those changes thanks!

feanil:
Ready for review again.

PR #427, opened 2019-08-07
Add additional extension points for use by Runtime/FieldData subclasses

bradenmacdonald:
Some minor changes that proved useful / necessary when building the Blockstore-based prototype XBlock runtime:

* Needed a way for FieldData implementations to know when `block.save()` was called, so that the field data implementation can persist all pending field data value changes in a single write, rather than multiple writes.
* Needed a way for `Runtime` subclasses to more easily add extra CSS classes to the `<div>` that wraps rendered XBlocks

*Still WIP*: Need to add tests and confirm no other changes are needed.

Impact: should be completely backwards compatible, doesn't change behavior.


xblock/runtime.py
```diff
@@ -919,6 +934,18 @@ def _wrap_ele(self, block, view, frag, extra_data=None):
         wrapped.add_fragment_resources(frag)
         return wrapped
 
+    def _css_classes_for(self, block, view):
+        """
+        Get the list of CSS classes that the wrapping <div> should have for the
+        spcified xblock or aside's view.
```
nedbat: "specified"

bradenmacdonald: I really need to find a better spell checker plugin for my editor, or something... thank you :)

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5393](https://openedx.atlassian.net/browse/OSPR-5393) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #428, opened 2019-09-18
Pass unicode to json.loads.

feanil:
In python2 there was auto-coercion happening here but in python3 we need
to pass in unicode and requests body is bytes.


xblock/mixins.py
```diff
@@ -61,7 +61,7 @@ def wrapper(self, request, suffix=''):
             if request.method != "POST":
                 return JsonHandlerError(405, "Method must be POST").get_response(allow=["POST"])
             try:
-                request_json = json.loads(request.body)
+                request_json = json.loads(request.body.decode('utf-8'))
```
cpennington: Nit: should this decode using the content type declared on the request?

feanil: Ended up just changing this to `request.text` so that the request object can manage encoding logic if it needs to.

feanil: Changed it back to this because, I'm not confident in the exact request object we are going to get and whether it will have that attribute.  Could be a request lib object or a django request or a wsgi request all of which are slightly different.  Gonna play it safe and be explicit.  Gonna just hardcode 'utf-8' here for the same reason.  I don't know how to get the charset out of the request object safely without making this really complicated.

cpennington: The contract for this is that it's a webob request object, but in practice it's actually a faked-up webob request built on top of a django request.

The Cale-of-6-years-ago was an overly ambitious sort.

PR #429, opened 2019-09-23
Add support for relative dates to DateTime fields

cpennington:


xblock/fields.py
```diff
@@ -965,29 +965,42 @@ def from_json(self, value):
...
+
             try:
                 value = dateutil.parser.parse(value)
             except (TypeError, ValueError):
                 raise ValueError("Could not parse {} as a date".format(value))
 
-        if not isinstance(value, datetime.datetime):
+        # Interpret raw numbers as a relative dates
+        if isinstance(value, (int, float)):
+            return datetime.timedelta(seconds=value)
```
jmyatt: this line might be better written as:
`value = datetime.timedelta(seconds=value)`

That way the general flow is 1) interpret/transform `value`, 2) validate `value`'s type 3) fix up and return `value`.

xblock/fields.py
```diff
@@ -965,29 +965,42 @@ def from_json(self, value):
...
         if isinstance(value, datetime.datetime):
             return value.strftime(self.DATETIME_FORMAT)
+
+        if isinstance(value, datetime.timedelta):
+            return value.total_seconds()
+
         if value is None:
             return None
+
         raise TypeError("Value stored must be a datetime object, not {}".format(type(value)))
```
jmyatt: maybe add timedelta as a possible stored value in this message?

xblock/fields.py
cpennington: Oh, yeah, the `return` was a brainfart.

bradenmacdonald:
Nice :)

PR #430, opened 2019-10-11
A faster sanitization of text

nedbat:
XML doesn't require us to remove as many characters as we were.  This sanitization appeared as a hot spot in profiling the LMS.  This makes it much faster (20x?)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py{27,36}-django111
+envlist = py{27,35,36}-django111
 
 [pytest]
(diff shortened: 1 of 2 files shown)


nedbat:
Should I do a release for edx-platform, or what's the process here?

doctoryes:
@nedbat You didn't bump the XBlock version - was that intentional?

ormsbee:
@nedbat: I think you just make another commit with the version bump and changelog entry, we thumbs that one, and after that merges, you tag it and push up the tag. Once the new version goes up on PyPI, edx-platform should pick up the update the next time it auto-updates dependencies.

ormsbee:
We seem to be missing a few changelog entries anyhow :(

PR #439, opened 2019-12-06
-bug-fix

Ayub-Khan:
Fixes a bug. https://github.com/edx/xblock-utils/pull/64


Ayub-Khan:
@jmbowman kindly review.

jmbowman:
`request.body` should always be bytes; I think the actual problem here is that some of the `xblock-utils` tests are incorrectly mocking the body to be a Unicode string instead.  For example: https://github.com/edx/xblock-utils/blob/master/tests/unit/test_publish_event.py#L36 (`json.dumps()` returns bytes under Python 2.7 but Unicode under Python 3).

PR #452, opened 2020-04-20
BOM-1537 : Upgrade to Python 3.8

mzulqarnain1:
Relevant JIRA : [https://openedx.atlassian.net/browse/BOM-1537](https://openedx.atlassian.net/browse/BOM-1537)

- Added Python 3.8 to testing
- Removed Django versions < 2.2
- Ran make upgrade
- Removed constraints no longer needed
- Fixed new quality failures

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -19,7 +19,6 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=caniusepython3.pylint_checker
 
 
 [MESSAGES CONTROL]
@@ -62,6 +61,7 @@ disable=
     too-many-locals,
     duplicate-code,
     len-as-condition,
+    fixme,
 
 
 [REPORTS]
--- a/setup.py
+++ b/setup.py
@@ -47,16 +47,10 @@ setup(
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Framework :: Django',
-        'Framework :: Django :: 1.11',
-        'Framework :: Django :: 2.0',
-        'Framework :: Django :: 2.1',
         'Framework :: Django :: 2.2',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
         'Natural Language :: English',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-        "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.5",
     ]
 )
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
         'Natural Language :: English',
+        "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.5",
     ]
 )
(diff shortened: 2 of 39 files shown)


tirkarthi:
@mzulqarnain1 See also #454 #453 for some potential warnings related to migration.

pylintrc
```diff
@@ -19,7 +19,7 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=caniusepython3.pylint_checker
+load-plugins=
```
awais786, line 22: @mzulqarnain1 there is nothing after `=`

xblock/django/request.py
```diff
@@ -146,15 +146,15 @@ def POST(self):
             querydict_to_multidict(self._request.FILES, wrap=DjangoUploadedFile),
         )
 
-    @lazy
-    def body(self):  # pylint: disable=method-hidden
+    @property
```
awais786, line 149: please add a reason for this change.

xblock/plugin.py
```diff
@@ -139,7 +138,7 @@ def load_classes(cls, fail_silently=True):
         """
         all_classes = itertools.chain(
             pkg_resources.iter_entry_points(cls.entry_point),
-            (entry_point for identifier, entry_point in cls.extra_entry_points),
+            (entry_point for identifier, entry_point in iter(cls.extra_entry_points)),
```
awais786, line 141: why you are doing this ?

xblock/runtime.py
```diff
@@ -54,22 +54,18 @@ def __new__(cls, scope, user_id, block_scope_id, field_name, block_family='xbloc
     @abstractmethod
     def get(self, key):
         """Reads the value of the given `key` from storage."""
```
awais786, line 56: I think you can disable the pass warning instead of removing.

xblock/django/request.py
mzulqarnain1: Both of these are marked as `property` in class we are inheriting from so pylint suggested making it `property` in child class too.

xblock/plugin.py
mzulqarnain1: This is a lazy member of our class which returns a list when called, but pylint doesn't consider it an iterable and warns, so wrapping it in `iter()` does the trick.

xblock/runtime.py
mzulqarnain1: Should I go on and re-add all the `pass`?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,35}-django111,py{35}-django{20,21,22}
+envlist = py{35,38}-django{22,30}
```
morenol, line 2: Django 3.0 with Python3.5 is an invalid combination

setup.py
```diff
@@ -47,16 +47,10 @@
...
-        'Framework :: Django :: 1.11',
-        'Framework :: Django :: 2.0',
-        'Framework :: Django :: 2.1',
         'Framework :: Django :: 2.2',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
         'Natural Language :: English',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-        "Programming Language :: Python :: 3",
```
jmbowman, line 59: We want to keep this one, some tools in the Python ecosystem use it to determine whether the package supports any version of Python 3.

PR #547, opened 2021-09-30
build: add a docs check to CI

nedbat:
(also, move the docs from "doc" to "docs" directory for uniformity with other repos.)


feanil:
Is there a reason you went with this instead of having RTD do the doc building on the PR based on a `.readthedocs.yml`?  I had started down this same path but ended up moving to that to make it more similar to how the docs would be built from master.

nedbat:
Updated to use a readthedocs pull-request build.  I removed the "docs" tox environment from the existing CI.  Ready for another review.

PR #550, opened 2021-10-08
fix build and update python requirements

kdmccormick:
See individual commit messages.


kdmccormick:
closed in favor of https://github.com/edx/XBlock/pull/545

PR #559, opened 2021-11-09
Standardize version number placement

aht007:
Description:
Standardize the placement of version number


docs/conf.py
```diff
@@ -10,21 +10,19 @@
...
 # serve to show the default.
 
-import codecs
 import datetime
 import os
 import sys
 
 import edx_theme
 from unittest import mock
+from ..xblock import __version__
```
aht007, line 19: @mraarif 
When I use relative import here the build fails raising an error that __name__ is. not is globals which I googled and found that we can't use relative imports in test files so I replaced it with "from XBlock.xblock" but then it raises error of "No module XBlock"
Any suggestions on it.....

mraarif (approved):

kdmccormick:
@aht007 Sorry, I won't be able to review any of these in the near future. Could you tag @edx/teaching-and-learning instead going forward?

PR #617, opened 2023-01-27
feanil/update django pyfs

feanil:
- feat: Update from django-pyfs to openedx-django-pyfs
- chore: Run `make upgrade`
- chore: Bump the version for the updated dependency.


Carlos-Muniz (approved):
LGTM!

Issue #676, opened 2023-05-25
Switch off of deprecated `pkg_resources` library


kdmccormick:
### Background

The `pkg_resources` library, used by XBlock to load static assets, [is deprecated](https://setuptools.pypa.io/en/latest/pkg_resources.html). The [removal timeline is unknown](https://github.com/pypa/setuptools/discussions/3934).

xblock-utils uses `pkg_resources` in a couple places.

### To do

First, choose a new resource loading interface. Options:

* Use [`importlib.resources`](https://docs.python.org/3.8/library/importlib.html?highlight=importlib#module-importlib.resources). Unfortunately, this will become deprecated in Python 3.11 and replaced with a yet-to-be-determined interface.
* Wait until Python 3.11, and then switch to the 3.11 replacement.

Then:

* Update the code in xblock-utils.

### Related

* https://github.com/openedx/XBlock/issues/641
* https://github.com/openedx/XBlock/issues/675

farhan:
After reading the related docs/blogs shared below I think we should choose [importlib_resources](https://pypi.org/project/importlib-resources/) library which is developed and maintained by Python developers and fixes the architecture issues related to [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources) module which is going to deprecated in Python 3.11 as discussed [here](https://discuss.python.org/t/deprecating-importlib-resources-legacy-api/11386)

Docs/Blogs
https://discuss.python.org/t/deprecating-importlib-resources-legacy-api/11386
https://importlib-resources.readthedocs.io/en/latest/using.html
https://docs.python.org/3/library/importlib.resources.html
https://www.geeksforgeeks.org/importlib-package-in-python/
https://dev.to/bowmanjd/easily-load-non-python-data-files-from-a-python-package-2e8g

FYI @kdmccormick @feanil

kdmccormick:
@farhan , thanks for the research. I took a look at the Python forum post and came to the conclusion that:

* `pkg_resources` is what we use now. It is deprecated but will still work in Python 3.11.
* `importlib.resources` was introduced in Python 3.7, but its API had design problems.
* `importlib_resources` is a better, redesigned version of that API.
* In Python 3.11, `importlib_resources` becomes official, as `importlib.resources`, thus deprecating the old, problematic version of the API.

So, it sounds like we could wait until edx-platform is upgraded to Python 3.11, and then just switch directly to the new, official `importlib.resources` package. Does that sound good?

feanil:
[...]

PR #716, opened 2024-02-06
Drop py3.8 support | Replace pkg_resources lib with importlib.resources

farhan:
- Drop Py3.8 support 
- Replace pkg_resources lib with importlib.resources

Ticket: https://github.com/openedx/XBlock/issues/676

**Followed migration guide:** 
https://importlib-resources.readthedocs.io/en/latest/migration.html

**Testing:**
I have tested the changes with following steps:
1. Use [PR](https://github.com/openedx/xblock-sdk/pull/350) of `xblock-sdk`
2. Install `xblock` into `xblock-sdk`

3. Install `feedback-xblock` from [PR](https://github.com/openedx/FeedbackXBlock/pull/58) into `xblock-sdk`
4. Test feedback xblock functionality

### DIFF
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -6,7 +6,6 @@ import functools
 import inspect
 import json
 import logging
-import os
 import warnings
 from collections import OrderedDict, defaultdict
 
@@ -161,8 +160,12 @@ class Blocklike(metaclass=_AutoNamedFieldsMetaclass):
 
     @classmethod
     def _open_resource(cls, uri):
-        return importlib.resources.files(inspect.getmodule(cls).__package__).joinpath(
-            os.path.join(cls.resources_dir, uri)
+        return importlib.resources.files(
+            inspect.getmodule(cls).__package__
+        ).joinpath(
+            cls.resources_dir
+        ).joinpath(
+            uri
         ).open('rb')
 
     @classmethod
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -10,7 +10,7 @@ import os
 import warnings
 from collections import OrderedDict, defaultdict
 
-import pkg_resources
+import importlib.resources
 from lxml import etree
 from webob import Response
 
@@ -157,7 +157,13 @@ class Blocklike(metaclass=_AutoNamedFieldsMetaclass):
         if "/." in uri:
             raise DisallowedFileError("Only safe file names are allowed: %r" % uri)
 
-        return pkg_resources.resource_stream(cls.__module__, os.path.join(cls.resources_dir, uri))
+        return cls.open_resource(uri)
+
+    @classmethod
+    def open_resource(cls, uri):
+        return importlib.resources.files(inspect.getmodule(cls).__package__).joinpath(
+            os.path.join(cls.resources_dir, uri)
+        ).open('rb')
 
     @classmethod
     def json_handler(cls, func):
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -157,10 +157,10 @@ class Blocklike(metaclass=_AutoNamedFieldsMetaclass):
         if "/." in uri:
             raise DisallowedFileError("Only safe file names are allowed: %r" % uri)
 
-        return cls.open_resource(uri)
+        return cls._open_resource(uri)
 
     @classmethod
-    def open_resource(cls, uri):
+    def _open_resource(cls, uri):
         return importlib.resources.files(inspect.getmodule(cls).__package__).joinpath(
             os.path.join(cls.resources_dir, uri)
         ).open('rb')
--- a/xblock/plugin.py
+++ b/xblock/plugin.py
@@ -4,9 +4,9 @@ Generic plugin support so we can find XBlocks.
 This code is in the Runtime layer.
 """
 import functools
+import importlib.metadata
 import itertools
 import logging
-import pkg_resources
 
 from xblock.internal import class_lazy
 
@@ -100,7 +100,11 @@ class Plugin:
             if select is None:
                 select = default_select
 
-            all_entry_points = list(pkg_resources.iter_entry_points(cls.entry_point, name=identifier))
+            all_entry_points = [
+                entry_point
+                for entry_point in importlib.metadata.entry_points(group=cls.entry_point)
+                if entry_point.name == identifier
+            ]
             for extra_identifier, extra_entry_point in iter(cls.extra_entry_points):
                 if identifier == extra_identifier:
                     all_entry_points.append(extra_entry_point)
@@ -133,7 +137,7 @@ class Plugin:
         contexts. Hence, the flag.
         """
         all_classes = itertools.chain(
-            pkg_resources.iter_entry_points(cls.entry_point),
+            importlib.metadata.entry_points(group=cls.entry_point),
             (entry_point for identifier, entry_point in iter(cls.extra_entry_points)),
         )
         for class_ in all_classes:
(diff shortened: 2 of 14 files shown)


kdmccormick:
@farhan per the conversation on https://github.com/openedx/XBlock/issues/676, could we convert this to a draft and pause it until after the Python 3.11/3.12 upgrade? That would allow us to avoid adding importlib_resources as a dependency.

feanil:
This looks good, but we're currently blocked on merging and releasing this because we have not yet dropped 3.8 support in edx-platform.  So let's hold off on merging for now.

farhan:
PR is blocked as we are waiting for drop of Python `v3.8` support from edx-platform

xblock/core.py
```diff
@@ -157,7 +157,13 @@ def open_local_resource(cls, uri):
         if "/." in uri:
             raise DisallowedFileError("Only safe file names are allowed: %r" % uri)
 
-        return pkg_resources.resource_stream(cls.__module__, os.path.join(cls.resources_dir, uri))
+        return cls.open_resource(uri)
+
+    @classmethod
+    def open_resource(cls, uri):
```
kdmccormick, line 163:
```suggestion
    def _open_resource(cls, uri):
```

I imagine that you extracted this into its own function for mocking purposes, right? That is all good, but by defining this method on a public class, we would be adding a new public method to the XBlock API.

Public API method exert a maintenance burden, because we must support them indefinitely. They can only be removed by announcing a DEPR, waiting for it to be accepted, and releasing a new major XBlock version. So, I would not add a new public method unless there is a strong reason to do so.

Instead, we could avoid that by marking the method as "private" with a leading underscore. That way, we could change or remove it in the future without breaking the public contract. How does that sound?

kdmccormick (changes requested):
Nice. Just a couple change requests. Glad to hear you tested this with xblock-sdk. I'm going to test this with edx-platform today or tomorrow, I'll let you know if I run into any issues.

xblock/core.py
farhan: That sounds great. 

I learned not to introduce public things into libraries unless absolutely necessary.

farhan:
@kdmccormick 
Requested changes addressed.
Do share the results of testing on `edx-platform`

farhan:
@kdmccormick Thanks for the testing of this important PR.

I have fixed the issues in this [comment] and tested on edx-platform (https://github.com/openedx/XBlock/pull/716/commits/aa6f2f9db41c5fb6dec262dbefeddca5541b92ce)

PR is ready for next pass.

Can you share some example course having different XBlocks on the latest `edx-platform`?

xblock/utils/resources.py
```diff
@@ -57,7 +56,8 @@ def render_mako_template(self, template_path, context=None):
         )
         context = context or {}
         template_str = self.load_unicode(template_path)
-        lookup = MakoTemplateLookup(directories=[pkg_resources.resource_filename(self.module_name, '')])
+        directory = os.path.dirname(os.path.realpath(sys.modules[self.module_name].__file__))
```
kdmccormick, line 59: Since `sys.modules` only includes modules that have already been imported, this will fail if `self.module_name` hasn't been imported yet, which is unlikely, but possible. So, I would use `importlib.import_module` just like you did in `load_unicode`, and then grab `__file__` from the result of that.

xblock/utils/resources.py
```diff
@@ -22,7 +22,7 @@ def load_unicode(self, resource_path):
         Gets the content of a resource
         """
         package_name = importlib.import_module(self.module_name).__package__
-        return importlib.resources.files(package_name).joinpath(resource_path).read_text()
+        return importlib.resources.files(package_name).joinpath(resource_path.lstrip('/')).read_text()
```
kdmccormick, line 25: Thanks for the `lstrip` fix, nice! It took me some time to understand the fix, though. I had to go into the lti_consumer block and notice that they are using absolute paths, which pkg_resources happily converted to a relative path, but which causes an exception with importlib, so you must strip off the leading slash yourself.

To help out future maintainers, please leave a brief comment explaining why you are stripping the leading slash here.

xblock/utils/resources.py
kdmccormick, line 59: I need to think about this line a bit. It makes me nervous that you are using direct path manipulation instead of the `importlib.resources` API, but I also understand that you probably chose to do that because the API doesn't give any good way to retrieve just the directory name, so path manipulation was the way to go. I'll mull it over and get back to you.

kdmccormick (changes requested):
A few more comments, but this is getting close. The Python packaging APIs drive me nuts... I appreciate you guys digging into this.

farhan:
@kdmccormick Thanks for such a quality review. Let me spend more time on it in depth

xblock/utils/resources.py
farhan: Actually I picked earlier line from [here](https://github.com/openedx/XBlock/blob/master/xblock/utils/resources.py#L90)

Update the code though based on importlib package, please review
I think `self.module_name` will always be imported as we are passing it in the `__init__` method of `ResourceLoader`

xblock/plugin.py
```diff
@@ -100,7 +100,7 @@ def select(identifier, all_entry_points):
             if select is None:
                 select = default_select
 
-            all_entry_points = list(pkg_resources.iter_entry_points(cls.entry_point, name=identifier))
+            all_entry_points = importlib.metadata.entry_points(group=cls.entry_point, name=identifier)
```
kdmccormick, line 103:
```suggestion
            all_entry_points = list(importlib.metadata.entry_points(group=cls.entry_point, name=identifier))
```

According to the docs this returns [a collection](https://docs.python.org/3/library/importlib.metadata.html#entry-points), which means it is not necessarily a list, but it can be safely converted to list.  Since we use it as a list (`all_entry_points.append`) a few lines down, we should ensure that it is in fact a list.

kdmccormick (approved):
Just two small comments, but if you like these, then this looks good to me! Thanks for your thorough work here.

I tested with the ProblemBlock, drag-and-drop-v2, and lti_consumer blocks. (Since you asked, the best course I know of for testing different XBlocks is the [openedx-test-course](https://github.com/openedx/openedx-test-course). It's not great, but it's better than nothing :)

xblock/utils/resources.py
kdmccormick: > Update the code though based on importlib package, please review

Nice, that looks a lot better to me.

> I think self.module_name will always be imported as we are passing it in the __init__ method of ResourceLoader

self.module_name is just a string; passing it to `__init__` does not guarantee that the module it points to has been parsed and loaded into `sys.modules` by the import system.

xblock/plugin.py
```diff
@@ -100,7 +100,11 @@ def select(identifier, all_entry_points):
             if select is None:
                 select = default_select
 
-            all_entry_points = list(pkg_resources.iter_entry_points(cls.entry_point, name=identifier))
+            all_entry_points = [
+                entry_point
+                for entry_point in importlib.metadata.entry_points(group=cls.entry_point)
+                if entry_point.name == identifier
+            ]
```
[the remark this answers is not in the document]
farhan: Tests are failing on Python 3.12, let me look into them on Monday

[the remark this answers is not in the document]
kdmccormick: Sounds great Farhan. Coincidentally I think that resolving [this comment](https://github.com/openedx/XBlock/pull/716#discussion_r1684582301) will fix up the tests.

Issue #714, opened 2024-01-31
[DEPR]: Extraneous XBlock Mixins


kdmccormick:
### Proposal Date

2024-01-31

### Target Ticket Acceptance Date

2024-02-09

### Earliest Open edX Named Release Without This Functionality

Redwood - 2024-04

### Rationale

The [implementations of the XBlock and XBlockAside classes](https://github.com/openedx/XBlock/blob/master/xblock/core.py#L126) are chopped into [about a dozen tiny mixins](https://github.com/openedx/XBlock/blob/master/xblock/mixins.py). This was done many years ago, probably for some mix of reasons including code organization and theoretical reusability. Now, in 2024 in the openedx GitHub org, there are uses of the XBlock, XBlockAside, and XBlockMixin classes, but we do not see uses of any of the other classes or mixins (other than one extraneous reference to HandlersMixin in edx-platform, which can be removed).

Unfortunately, these mixins make it harder/impossible to write robust type constraints for XBlock and XBlockAside. They also make it harder to find code in the XBlock package. [See discussion](https://github.com/openedx/XBlock/issues/707#issuecomment-1919915301).

### Removal

In an upcoming XBlock 2.0.0 release...

* **Removed:** class `xblock.core.SharedBlockBase`.
* **Removed:** module `xblock.mixins`, including...
  * these classes, collapsed into `xblock.core.Blocklike`:
    * `xblock.mixins.RuntimeServicesMixin`
    * `xblock.mixins.ScopedStorageMixin`
    * `xblock.mixins.IndexInfoMixin`
    * `xblock.mixins.XmlSerializationMixin`
    * `xblock.mixins.HandlersMixin`
  * these classes, collapsed into `xblock.core.XBlock`:
    * `xblock.mixins.ChildrenModelMetaclass`
    * `xblock.mixins.HierarchyMixin`
    * `xblock.mixins.ViewsMixin`
* **New:** class `xblock.core.Blocklike`, serving as a shared base for `xblock.core.XBlockMixin`, `xblock.core.XBlock`, and `xblock.core.XBlockAsidse`.
* **No functional changes** to `xblock.plugin.Plugin`, `xblock.core.XBlock` and `xblock.core.XBlockAside`.
* **Backwards-compatible enhancements** to `xblock.core.XBlockMixin` include the functionality of these former classes:
    * `xblock.mixins.IndexInfoMixin`
    * `xblock.mixins.XmlSerializationMixin`
[...]

Issue #707, opened 2024-01-12
Add type hints and check them in CI


kdmccormick:
XBlock is a widely used core library, but is pretty abstract and thus can be hard to reason about. So, this would be a particularly valuable repo to add type hinting to. 

Some additional notes are here:

* https://github.com/openedx/XBlock/issues/710

Draft PR here:

* https://github.com/openedx/XBlock/pull/761

kdmccormick:
early WIP: https://github.com/openedx/XBlock/pull/713

kdmccormick:
@bradenmacdonald I've been working on this, and it's going pretty well. I'm actually really impressed how far mypy has come--typing XBlock Fields as generic descriptors works great!

The hardest part so far has been dealing with how XBlock's implementation is chopped up into a bunch of tiny mixins. I guess that was in anticipation of supporting a bunch of XBlock-like classes that don't use all of XBlock's capability, but AFAICT we only have XBlock and XBlockAside. A good demonstration of the pain is HierarchyMixin, which is written in such a way that I can't see it ever working outside of the XBlock class. [Here's the typing hack I'm using to get around it](https://github.com/openedx/XBlock/pull/713/files#diff-1585c15f188d01fc631340669749dde57fcb8c3203b1c95eecf58ee0d69ffa3fR390-R397).

I'm really tempted to collapse all the mixins down into a simple hierarchy:
```python
class Plugin:
    ... # as is
class ScopedStorageMixin:
    ... # as is

class XBlockMixin(ScopedStorageMixin):
    ... # as is

class SharedBlockBase(ScopedStorageMixin, Plugin):
    ... # collapse in: XmlSerializationMixin, HandlersMixin

class XBlockAside(SharedBlockBase):
    ... # as is

class XBlock(SharedBlockBase):
    ... # collapse in: HierarchyMixin, ViewsMixin
```

Does that sound like a good move? Or do you think the current set of mixins is worth keeping around?

kdmccormick:
Alternatively, here's an even more aggressive simplification. It would add capability to XBlockMixin, but I honestly think we'll need that extra capability when we go to apply type-checking to the XBlockMixins in edx-platform:
```python
class Plugin:
   ... # as-is
class Blocklike:
   ... # collapse in: RuntimerServicesMixin, ScopedStorageMixin, XmlSerializationMixin, HandlersMixin

class XBlockMixin(Blocklike):
   ... # as is

class XBlockAside(Blocklike, Plugin):
[...]

PR #718, opened 2024-02-12
feat!: collapse extraneous XBlock mixins

kdmccormick:
## Supporting info

Proposed in this DEPR issue:

* https://github.com/openedx/XBlock/issues/714

excerpt:
> The [implementations of the XBlock and XBlockAside classes](https://github.com/openedx/XBlock/blob/master/xblock/core.py#L126) are chopped into [about a dozen tiny mixins](https://github.com/openedx/XBlock/blob/master/xblock/mixins.py). This was done many years ago, probably for some mix of reasons including code organization and theoretical reusability. Now, in 2024 in the openedx GitHub org, there are uses of the XBlock, XBlockAside, and XBlockMixin classes, but we do not see uses of any of the other classes or mixins (other than one extraneous reference to HandlersMixin in edx-platform, which can be removed).

> Unfortunately, these mixins make it harder/impossible to write robust type constraints for XBlock and XBlockAside. They also make it harder to find code in the XBlock package. [See discussion](https://github.com/openedx/XBlock/issues/707#issuecomment-1919915301).

The edx-platform upgrade PR is here:

* https://github.com/openedx/edx-platform/pull/34231

## Description

Various extraneous classes have been removed from the XBlock API. We believe that most, if not all, XBlock API users will be unaffected by this change.

### Diagram

Generated with `pyreverse` and `dot`.

#### Before

xblock-before

#### After

xblock-after

### Change list

* **Removed:** 
  * `xblock.XBlockMixin` (still available as `xblock.core.XBlockMixin`)
  * `xblock.core.SharedBlockBase` (replaced with `xblock.core.Blocklike`)
  * `xblock.internal.Nameable`
  * `xblock.internal.NamedAttributesMetaclass`
  * `xblock.django.request.HeadersDict`
  * `xblock.fields.XBlockMixin` (still available as `xblock.core.XBlockMixin`)
  * `xblock.mixins.RuntimeServicesMixin`
  * `xblock.mixins.ScopedStorageMixin`
  * `xblock.mixins.IndexInfoMixin`
  * `xblock.mixins.XmlSerializationMixin`
  * `xblock.mixins.HandlersMixin`
  * `xblock.mixins.ChildrenModelMetaclass`
  * `xblock.mixins.HierarchyMixin`
  * `xblock.mixins.ViewsMixin`
* **Added:**
  * `xblock.core.Blocklike`, the new common ancestor of `XBlock` and `XBlockAside`, and `XBlockMixin`, replacing `xblock.core.SharedBlockBase`.
  * New attributes on `xblock.core.XBlockAside`, each behaving the same as their `XBlock` counterpart:
    * `usage_key`
    * `context_key`
    * `index_dictionary`
  * Various new attributes on `xblock.core.XBlockMixin`, encompassing the functionality of these former classes:
    * `xblock.mixins.IndexInfoMixin`
    * `xblock.mixins.XmlSerializationMixin`
    * `xblock.mixins.HandlersMixin`
* **Docs**
  * Various docstrings have been improved, some of which are published in the docs.
  * XBlockAside will now be represented in the API docs, right below XBlock on the "XBlock API" page.
  * XBlockMixin has been removed from the docs. It was only ever documented under the "Fields API" page (which didn't make any sense), and it was barely even documented there. We considered adding it back to the "XBlock API"  page, but as noted in the class's new docstring, we do not want to encourage any new use of XBlockMixin.

Bumps version from 2.0.0 to 3.0.0.

### Docs

You can preview the doc additions by clicking "Details" on the docs build, hitting "View Docs", and going to the "XBlock API" page).

Homepage: https://docsopenedxorg--718.org.readthedocs.build/projects/xblock/en/718/index.html
New XBlockAside generated docs: https://docsopenedxorg--718.org.readthedocs.build/projects/xblock/en/718/xblock.html#xblock.core.XBlockAside

### XBlock and XBlockAside attributes

From the root of the repo, you can run this script in order to compare the list of attributes on XBlock/XBlockAside before and after this PR, both for the classes and their instances.
```bash

rm -rf dir_before dir_after
mkdir dir_before dir_after

cd dir_before
git switch master
python -c 'from xblock import core; print(*sorted(dir(core.XBlock)), sep="\n")' > block_cls.list
python -c 'from xblock import core; print(*sorted(dir(core.XBlock(None, None, None))), sep="\n")' > block_obj.list
python -c 'from xblock import core; print(*sorted(dir(core.XBlockAside)), sep="\n")'  > aside_cls.list
python -c 'from xblock import core; print(*sorted(dir(core.XBlockAside(None, None, runtime=None))), sep="\n")' > aside_obj.list
cd ..

cd dir_after
git switch kdmccormick/mixins
python -c 'from xblock import core; print(*sorted(dir(core.XBlock)), sep="\n")' > block_cls.list
python -c 'from xblock import core; print(*sorted(dir(core.XBlock(None, None, None))), sep="\n")' > block_obj.list
python -c 'from xblock import core; print(*sorted(dir(core.XBlockAside)), sep="\n")'  > aside_cls.list
python -c 'from xblock import core; print(*sorted(dir(core.XBlockAside(None, None, runtime=None))), sep="\n")' > aside_obj.list
cd ..

diff dir_before list dir_after
```

This should yield:
```
diff dir_before/aside_cls.list dir_after/aside_cls.list
39a40
> context_key
49a51
> index_dictionary
62a65
> usage_key
diff dir_before/aside_obj.list dir_after/aside_obj.list
42a43
> context_key
52a54
> index_dictionary
67a70
> usage_key
diff dir_before/block_cls.list dir_after/block_cls.list
0a1
> __annotations__
diff dir_before/block_obj.list dir_after/block_obj.list
0a1
> __annotations__
```

which indicates that:
* XBlockAsides now have `context_key`, `usage_key` and `index_dictionary` attributes
* XBlocks have a few type annotations now (more to come in #707 )
* The list of attributes is otherwise unchanged


docs/conf.py
```diff
@@ -128,6 +128,7 @@
     ('py:class', 'aside'),
     ('py:class', 'aside_fn'),
     ('py:class', 'webob.Request'),
+    ('py:class', 'webob.Response'),
```
kdmccormick, line 131: Added so that we can reference the "Response" class in a new docstring.

xblock/__init__.py
```diff
@@ -2,29 +2,4 @@
...
-    Deprecated.
-    """
-    def __init__(self, *args, **kwargs):
-        warnings.warn("Please use xblock.core.XBlockMixin", DeprecationWarning, stacklevel=2)
-        super().__init__(*args, **kwargs)
-
-
-# For backwards compatibility, provide the XBlockMixin in xblock.fields
-# without causing a circular import
-xblock.fields.XBlockMixin = XBlockMixin
```
kdmccormick, line 28: We're just removing the wrapper locations, which have been deprecated since 2014. XBlockMixin is still available at xblock.core.XBlockMixin.

xblock/core.py
```diff
@@ -1,83 +1,121 @@
...
 
-# exposing XML_NAMESPACES as a member of core, in order to avoid importing mixins where
-# XML_NAMESPACES are needed (e.g. runtime.py).
-XML_NAMESPACES = xblock.mixins.XML_NAMESPACES
+# OrderedDict is used so that namespace attributes are put in predictable order
+# This allows for simple string equality assertions in tests and have no other effects
+XML_NAMESPACES = OrderedDict([
+    ("option", "http://code.edx.org/xblock/option"),
+    ("block", "http://code.edx.org/xblock/block"),
+])
```
kdmccormick, line 34: Comment copied in from xblock.mixins

xblock/core.py
```diff
@@ -1,83 +1,121 @@
...
 
 # __all__ controls what classes end up in the docs.
-__all__ = ['XBlock']
+__all__ = ['XBlock', 'XBlockAside']
```
kdmccormick, line 37: This puts XBlockAside in the docs

xblock/core.py
```diff
@@ -1,83 +1,121 @@
...
+
 UNSET = object()
 
 
-class XBlockMixin(ScopedStorageMixin):
+class _AutoNamedFieldsMetaclass(type):
```
kdmccormick, line 42: xblock.internal.NamedAttributesMetaclass was responsible for sticking `__name__` attributes on any `Nameable` objects defined on its classes. In practice, Field was the only type of Nameable attribute.

_AutoNamedFieldsMetaclass is essentially the same as NamedAttributesMetaclass, but to simplify things, it only just operates on Fields.

_AutoNamedFieldsMetaclass continues to be the metaclass for both XBlock and XBlockAside (although XBlock further specializes it as _HasChildrenMetaclass, defined further down).

xblock/core.py
```diff
@@ -1,83 +1,121 @@
...
     """
+    Shared base for XBlocks and XBlockAsides, providing these common capabilities:
+
+    - services
+    - fields
+    - OLX
+    - handlers
+    - resources
 
+    (see XBlock and XBlockAside classes for details)
```
kdmccormick, line 92: Blocklike is our replacement for SharedBlockBase, plus all the capabilities that were shared by XBlock and XBlockAside anyway. Each capability mentioned in the docstring maps to a collapsed-in mixin, except for "resources", which were defined right on SharedBlockBase).

xblock/core.py
```diff
@@ -1,83 +1,121 @@
...
+    """
     resources_dir = ''
     public_dir = 'public'
     i18n_js_namespace = None
 
     @classmethod
     def get_resources_dir(cls):
         """
-        Gets the resource directory for this XBlock.
+        Gets the resource directory for this XBlock-like class.
```
kdmccormick, line 101: I've tried to generalize these docstrings because they apply to both XBlocks and XBlockAsides. Hopefully "XBlock-like class" isn't too clunky for folks reading the API docs.

xblock/core.py
```diff
@@ -121,23 +159,535 @@ def open_local_resource(cls, uri):
...
+    (...or don't do that, because it's confusing for other developers, and linters/type-checkers
+        won't understand that the mixins are part of XBlock instances, so your code will be full
+        of ignore statements :)
+    """
+
+
+class _HasChildrenMetaclass(_AutoNamedFieldsMetaclass):
+    """
+    Adds a ``children`` XBlock ReferenceList field to classes where ``has_children == True``.
+    """
```
kdmccormick, line 647: _HasChildrenMetaclass is the exact same as ChildrenModelMetaclass, but I've renamed it over here to be private and less abstract-sounding.

Side note, jfc this whole metaclass is such a terrible hack. It'll be totally impossible to make mypy understand it. Why couldn't we have just put `children` on a subclass of XBlock?
```python
class XBlockWithChildren(XBlock):
    children = ReferenceList(...)

    @property
    def has_children(self):
        return True
```

xblock/core.py
```diff
@@ -202,34 +802,18 @@ def __init__(self, runtime, field_data=None, scope_ids=UNSET, *args, **kwargs):
         if scope_ids is UNSET:
             raise TypeError('scope_ids are required')
 
-        # Provide backwards compatibility for external access through _field_data
-        super().__init__(runtime=runtime, scope_ids=scope_ids, field_data=field_data, *args, **kwargs)
-
-    @property
-    def usage_key(self):
```
kdmccormick, line 209: usage_key and context_key are moved to Blocklike so that XBlockAsides can use them too.

xblock/fields.py
```diff
@@ -339,6 +339,13 @@ def default(self):
         else:
             return self._default
 
+    @staticmethod
+    def needs_name(field):
+        """
+        Returns whether the given ) is yet to be named.
+        """
+        return not field.__name__
+
```
kdmccormick, line 348: This was copied in from Nameable, but I recall now that I've refactored _AutoNameFieldsMetaclass such that this is now unused. I'll remove it before merging.

```suggestion
```

xblock/internal.py
```diff
@@ -28,42 +27,3 @@ def __get__(self, instance, owner):
 
 
 class_lazy = LazyClassProperty  # pylint: disable=invalid-name
-
-
-class NamedAttributesMetaclass(type):
```
kdmccormick, line 33: NamedAttributesMetaclass and Nameable have been de-generalized into xblock.core._AutoNamedFieldsMetaclass and xblock.fields.Field, respectively.

xblock/test/test_core.py
```diff
@@ -435,8 +435,9 @@ class FieldTester(XBlock):
 
 def test_object_identity():
     # Check that values that are modified are what is returned
-    class FieldTester(ScopedStorageMixin):
-        """Toy class for ModelMetaclass and field access testing"""
+    @Blocklike.needs("field-data")
+    class FieldTester(Blocklike):
```
kdmccormick, line 439: RuntimeServicesMixin defined the `@needs` decorator, which allowed its subclass ScopedStorageMixin to invoke `@needs("field-data")`.

However, now Blocklike defines `@needs`, so Blocklike itself cannot invoke `@needs("field-data")`.

Thus, XBlock and XBlockAside now each invoke `@needs("field-data")` individually, and thus tests that subclass Blocklike must also invoke `@needs("field-data")`.

xblock/test/test_parsing.py
```diff
@@ -344,7 +344,7 @@ def test_dict_and_list_export_format(self):
     def test_unknown_field_as_attribute_raises_warning(self, parameter_name):
         with mock.patch('logging.warning') as patched_warn:
             block = self.parse_xml_to_block(f"<leaf {parameter_name}='something irrelevant'></leaf>")
-            patched_warn.assert_called_once_with("XBlock %s does not contain field %s", type(block), parameter_name)
+            patched_warn.assert_called_once_with("%s does not contain field %s", type(block), parameter_name)
```
kdmccormick, line 347: This warning was incorrect when the object was an XBlockAside.

I considered changing it to "Blocklike", but then the message would be:

`Blocklike FooBlock does not contain field bar`

which isn't any better than:

`FooBlock does not contain field bar`.

I could be convinced otherwise, esp if folks feel that the former is more regex-able.

kdmccormick (commented):
Notes for reviewers 👀

xblock/core.py
```diff
@@ -121,23 +159,530 @@ def open_local_resource(cls, uri):
...
+                # Only set the value if forced via xml_node;
+                # in all other cases, the value is None.
+                # Avoids an unnecessary XML end tag.
+                elem.text = text_value
+        else:
+            # Field will be output to XML as an attribute on the node.
+            node.set(field_name, text_value)
+
+
+class XBlockMixin(Blocklike):
```
ormsbee, line 626: If `XBlockMixin` is a subclass of `Blocklike` that overrides no other behavior, why not rename `Blocklike` to `XBlockMixin`? That would be more in line with our conventions, wouldn't it (where `XBlock` and `XBlockAside` both inherit from `XBlockMixin`)?

xblock/core.py
ormsbee: Well, children are mostly only useful in the edx-platform bundled XBlocks, so a DEPR and refactoring is probably feasible if you really want to kill this...

xblock/test/test_core.py
ormsbee: Can you manually invoke the decorator on `Blocklike` after it's defined? I haven't done something like this in ages, but I think it'd be something like this?

```python
class Blocklike:
    # lots of stuff here

Blocklike = Blocklike.needs(Blocklike, "field-data")
```

ormsbee (commented):
Just some comments/questions. No major concerns.

xblock/core.py
kdmccormick: So XBlockMixin is an unfortunate name. "Mixin" in the traditional sense means a class that isn't a meaningful base in itself, but is to be "mixed in" to other classes to add extra functionality. In that sense, the XBlockMixin isn't a mixin; rather, it's supposed to be a *base class for new custom XBlock mixins*. 

Contrast this with Blocklike, which is a meaningful base, one which I expect to be used all over xblock and edx-platform as a type annotation:

```python
def this_function_could_take_a_block_or_an_aisde(block: Blocklike):
    ...
```

...which I think reads better than using XBlockMixin as the annotation:

```python
def this_function_could_take_a_block_or_an_aisde(block: XBlockMixin):
    ...
```

I guess we could still reconcile the two classes by just aliasing `XBlockMixin = Blocklike`... what do you think @ormsbee ?

xblock/core.py
```diff
@@ -121,23 +159,530 @@ def open_local_resource(cls, uri):
...
     """
+    Base class for custom XBlock mixins.
+
+    To provide custom attributes to all XBlock instances in a Runtime, extend this class and
+    supply it to the Runtime's `mixins` parameter.
+
+    (...or don't do that, because it's confusing for other developers, and linters/type-checkers
+        won't understand that the mixins are part of XBlock instances, so your code will be full
+        of ignore statements :)
```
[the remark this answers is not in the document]
kdmccormick: @ormsbee You know, on second thought, what I really want is for people to avoid adding new xblock mixins to the *edx-platform* runtime. If people want to add new xblock mixins to their custom fork or custom runtime, that's A-OK. I'll move my big DO-NOT-USE-THIS comment to edx-platform instead.

xblock/core.py
kdmccormick: That is a _great_ point. It would also move us towards our desired end state where XBlocks are leaf-only.

kdmccormick: * https://github.com/openedx/XBlock/issues/729

xblock/core.py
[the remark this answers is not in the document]
kdmccormick: https://github.com/openedx/edx-platform/pull/34231/commits/aa8672b5a7cb784eac2fe0ec15807f4bd4c56798#diff-a16fd1c6cf75bdebdf97a2633b533d39e43975c6e98af58787ed6af1282eca71R980-R984

kdmccormick:
@ormsbee 
> Can you manually invoke the decorator on Blocklike after it's defined? I haven't done something like this in ages, but I think it'd be something like this?

Good idea. That bit of backwards incompatibility was making me a little nervous, so I'm glad to smooth it out. The syntax ended up being:
```python
Blocklike.needs('field-data')(Blocklike)
```

xblock/core.py
ormsbee: Okay, you've convinced me. Based on your description, `XBlockBase` would also work for me, but I'm good with `Blocklike`. I do like the idea of the alias since I have no idea if someone might have built off of that in their fork/plugin.

kdmccormick: @ormsbee On second thought, I'm actually inclined to leave XBlockMixin as its own subclass of Blocklike. It can't put my finger on exactly why, but it seems like it'd be *less surprising* to have XBlockMixin be its own class rather than an alias. Like you say, XBlockMixin is part of the public API, so we can't just delete it.

ormsbee (approved):

Issue #774, opened 2024-07-28
old and incomplete docs


CodeWithEmad:
XBlocks are the heart of Open edX. Yet, somehow the documentation around it is not complete.
There are a lot of places that are referenced for future fixes or mentioned TODO. like:

- There's no information about Runtime services. 
  https://docs.openedx.org/projects/xblock/en/latest/runtime.html#xblock.runtime.Runtime.service
  I found this in Jira but it's quite old: https://openedx.atlassian.net/wiki/spaces/AC/pages/161400730/Open+edX+Runtime+XBlock+API
- https://docs.openedx.org/projects/xblock/en/latest/xblock-tutorial/introduction.html#other-open-edx-resources
  > You, the user community, can help update and revise this documentation project on GitHub.
- Nowhere inside the docs is mentioned about [cookiecutter-xblock](https://github.com/openedx/edx-cookiecutters/tree/master/cookiecutter-xblock), but I think that's the standard approach to start with xblocks.
  https://edx.readthedocs.io/projects/xblock/en/latest/xblock-tutorial/getting_started/
- and many more

PS. we can add a tutor approach for installing xblocks. 

@openedx/axim-engineering

CodeWithEmad:
Also, the URL in the repository about section return 404
https://edx.readthedocs.io/projects/xblock-tutorial/en/latest/overview/introduction.html -> should be https://docs.openedx.org/projects/xblock/en/latest/index.html

sarina:
Hi @CodeWithEmad - is this a ticket you're planning to tackle, and would like some advice? Or are you making a general ticket for someone else to work on?

CodeWithEmad:
Hey @sarina. I am interested in contributing to this project, but I am not fully current with the recent Xblocks changes. It would be very beneficial to have a brief discussion with someone who is more closely involved with Xblock-related matters.

sarina:
@CodeWithEmad ah sorry that's not me 😅 

I think some of it you could get started on like adding notes about the cookiecutter. This seems like a big ticket that would probably benefit from multiple PRs to make smaller changes to get feedback on. You could also try posting specific questions in the forum (maybe one at a time, to generate a more focused discussion)

kdmccormick:
@CodeWithEmad You are right, the XBlock docs could really use some work, and your suggestions are good ones. +1 to Sarina's suggestion with breaking of proposal into smaller PRs, and the cookiecutter note is definitely a safe place to start.

[...]

PR #778, opened 2024-08-07
feat: XBlock overrides

### LINKED ISSUE #777: Xblock version 5.0.0 broke plugin
We have a plugin that is replacing the default XBlock with our own implementation that was recently broken with what we believe was the removal of `pkg_resources` support here [https://github.com/openedx/XBlock/pull/716](https://github.com/openedx/XBlock/pull/716).

We followed this post as our way of implementation [https://discuss.openedx.org/t/override-default-edx-platform-xblock-mappings-without-forking/12441](https://discuss.openedx.org/t/override-default-edx-platform-xblock-mappings-without-forking/12441)

Suspected root of the issue in our plugin:
```
    def ready(self):
        """
        Django startup code: swap edx-platform mapping for blocks with our translated versions.

        See https://discuss.openedx.org/t/override-default-edx-platform-xblock-mappings-without-forking/12441
        """

        # List of translated block types, for dismbiguating entry points
        translated_blocks = ["html", "problem", "video"]

        for block_type in translated_blocks:
            block_entry_points = list(
                pkg_resources.iter_entry_points("xblock.v1", block_type)
            )

            # If we have multiple entry points, ours gets precedence over the Open-edX copy
            if len(block_entry_points) > 1:
                dist = pkg_resources.get_distribution("Open-edX")
                del dist.get_entry_map("xblock.v1")[block_type]
```

nsprenkle:
Add ability to override another XBlock with the `xblock.v1.overrides` entrypoint. 

## Notes

- Fixes #777 which removed our current hack for overriding existing XBlock behavior with a plugin.
- See [conversation here](https://discuss.openedx.org/t/override-default-edx-platform-xblock-mappings-without-forking/12441/) for historical context & reasoning.

## Usage

``` python
setup(
    # ...
    entry_points={
        "xblock.v1.overrides": [
            "html = my_new_repo.my_new_block:HtmlBlock",
            "problem = my_new_repo.my_new_block:ProblemBlock",
            "video = my_new_repo.my_new_block:VideoBlock",
        ],
    }
)
```

## Tests

- [x] Unit tests
- [x] Smoke tests in LMS
  - [x] Overridden block types work as expected
  - [x] Non-overridden block types work as expected
- [x] Tests in Studio
  - [x] Overridden block types work as expected
  - [x] Non-overridden block types work as expected
- [x] Import tests
  - [x] Overridden block types work as expected
  - [x] Non-overridden block types work as expected
- [x] Export tests
  - [x] Overridden block types work as expected
  - [x] Non-overridden block types work as expected
- [ ] Other????

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,8 @@ Unreleased
 5.1.0 - 2024-08-07
 ------------------
 
-* added ability to override an XBlock with the 'xblock.{version}.overrides' entry point.
+* added ability to override an XBlock with the 'xblock.v1.overrides' entry point
+* added ability to override an XBlock Aside with the 'xblock_asides.v1.overrides' entry point
 
 
 5.0.0 - 2024-05-30
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,12 @@ Change history for XBlock
 Unreleased
 ----------
 
+5.1.0 - 2024-08-07
+------------------
+
+* added ability to override an XBlock with the 'xblock.{version}.overrides' entry point.
+
+
 5.0.0 - 2024-05-30
 ------------------
 
--- a/docs/xblock-tutorial/edx_platform/overrides.rst
+++ b/docs/xblock-tutorial/edx_platform/overrides.rst
@@ -9,7 +9,7 @@ implementation.
 
 This can be done by:
 
-1. Creating an XBlock in a new or existing Django Plugin, installed into ``edx-platform``.
+1. Creating an XBlock in a new or existing package installed into ``edx-platform``.
 
 2. Adding the ``xblock.v1.overrides`` entry point in ``setup.py``, pointing to your
    custom XBlock.
@@ -28,8 +28,8 @@ Imagine there is an XBlock installed ``edx-platform``:
 
 .. code:: python
 
-    # edx-platform/xblocks/twitch_plays_pokemon.py
-    class TwitchPlaysPokemonBlock(XBlock):
+    # edx-platform/xblocks/video_block.py
+    class VideoBlock(XBlock):
     ...
 
     # edx-platform/setup.py
@@ -38,18 +38,18 @@ Imagine there is an XBlock installed ``edx-platform``:
 
         entry_points={
             "xblock.v1": [
-                "tpp = xblocks.twitch_plays_pokemon::TwitchPlaysPokemonBlock"
+                "video = xblocks.video_block::VideoBlock"
                 # ...
             ]
         }
     )
 
-If you then create your own Plugin App with a custom version of that XBlock...
+If you then create your own Python package with a custom version of that XBlock...
 
 .. code:: python
 
-    # your_plugin/xblocks/twitch_plays_pokemon.py
-    class YourTwitchPlaysPokemonBlock(XBlock):
+    # your_plugin/xblocks/video_block.py
+    class YourVideoBlock(XBlock):
     ...
 
     # your_plugin/setup.py
@@ -57,14 +57,14 @@ If you then create your own Plugin App with a custom version of that XBlock...
         # ...
         entry_points={
             "xblock.v1.overrides": [
-                "tpp = your_plugin.twitch_plays_pokemon::YourTwitchPlaysPokemonBlock"
+                "tpp = your_plugin.xblocks.video_block::YourVideoBlock"
                 # ...
             ],
         }
     )
 
-And install that app as a Django plugin, then your block should be loaded instead of the
-existing implementation.
+And install that package into your virtual environment, then your block should be
+loaded instead of the existing implementation.
 
 .. note::
 
--- a/docs/xblock-tutorial/edx_platform/overrides.rst
+++ b/docs/xblock-tutorial/edx_platform/overrides.rst
@@ -1,0 +1,75 @@
+.. _Replace a Preinstalled XBlock With a Custom Implementation:
+
+##########################################################
+Replace a Preinstalled XBlock With a Custom Implementation
+##########################################################
+
+In XBlock ``v5.1.0``, the ability was introduced to override an XBlock with a custom
+implementation.
+
+This can be done by:
+
+1. Creating an XBlock in a new or existing Django Plugin, installed into ``edx-platform``.
+
+2. Adding the ``xblock.v1.overrides`` entry point in ``setup.py``, pointing to your
+   custom XBlock.
+
+This works with updated logic in ``load_class``'s ``default_select``, which gives
+load priority to a class with the ``.overrides`` suffix.
+
+This can be disabled by providing a different ``select`` kwarg to ``load_class`` which
+ignores or otherwise changes override logic.
+
+*******
+Example
+*******
+
+Imagine there is an XBlock installed ``edx-platform``:
+
+.. code:: python
+
+    # edx-platform/xblocks/twitch_plays_pokemon.py
+    class TwitchPlaysPokemonBlock(XBlock):
+    ...
+
+    # edx-platform/setup.py
+    setup(
+        # ...
+
+        entry_points={
+            "xblock.v1": [
+                "tpp = xblocks.twitch_plays_pokemon::TwitchPlaysPokemonBlock"
+                # ...
+            ]
+        }
+    )
+
+If you then create your own Plugin App with a custom version of that XBlock...
+
+.. code:: python
+
+    # your_plugin/xblocks/twitch_plays_pokemon.py
+    class YourTwitchPlaysPokemonBlock(XBlock):
+    ...
+
+    # your_plugin/setup.py
+    setup(
+        # ...
+        entry_points={
+            "xblock.v1.overrides": [
+                "tpp = your_plugin.twitch_plays_pokemon::YourTwitchPlaysPokemonBlock"
+                # ...
+            ],
+        }
+    )
+
+And install that app as a Django plugin, then your block should be loaded instead of the
+existing implementation.
+
+.. note::
+
+    The ``load_class`` code will throw an error in the following cases:
+
+    1. There are multiple classes attempting to override one XBlock implementation.
+
+    2. There is an override provided where an existing XBlock implementation is not found.
--- a/docs/xblock-tutorial/edx_platform/overrides.rst
+++ b/docs/xblock-tutorial/edx_platform/overrides.rst
@@ -57,7 +57,7 @@ If you then create your own Python package with a custom version of that XBlock.
         # ...
         entry_points={
             "xblock.v1.overrides": [
-                "tpp = your_plugin.xblocks.video_block::YourVideoBlock"
+                "video = your_plugin.xblocks.video_block::YourVideoBlock"
                 # ...
             ],
         }
--- a/xblock/plugin.py
+++ b/xblock/plugin.py
@@ -52,7 +52,7 @@ def select_with_overrides(identifier, all_entry_points):
     block_entry_points = []
 
     for block_entry_point in all_entry_points:
-        if block_entry_point.group.endswith('overrides'):
+        if block_entry_point.group.endswith('.overrides'):
             overrides.append(block_entry_point)
         else:
             block_entry_points.append(block_entry_point)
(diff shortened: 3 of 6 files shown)


farhan (changes requested):
@nsprenkle Thanks for creating the PR
- Didn't do the manual testing
- Done with the quick review, suggest some changes
- Was thinking we are packing then un-packing the entrypoints [here](https://github.com/openedx/XBlock/pull/778/files#diff-c9f31d34de1805791e8520e4ba2fe7c99e14be9f4f490b6d8bfc346b9aede117R55) but to keep the signature of method seems alright to me

nsprenkle:
@farhan, thanks for the review. Have tried to address all your comments and is ready for your re-review

xblock/plugin.py
```diff
@@ -41,6 +41,32 @@ def default_select(identifier, all_entry_points):  # pylint: disable=inconsisten
         raise AmbiguousPluginError(all_entry_points)
 
 
+def select_with_overrides(identifier, all_entry_points):
```
farhan, line 44: I think we should keep this method private to avoid its availability to library users

farhan:
@kdmccormick PR seems alright to me. Your review is mandatory.

xblock/plugin.py
kdmccormick: +1

xblock/plugin.py
```diff
@@ -41,6 +41,32 @@ def default_select(identifier, all_entry_points):  # pylint: disable=inconsisten
...
+    Honors the ability for an XBlock to override the default entry point.
+    Raise an exception when we have ambiguous entry points.
+    """
+
+    # Split entry points into overrides and non-overrides
+    overrides = []
+    block_entry_points = []
+
+    for block_entry_point in all_entry_points:
+        if block_entry_point.group.endswith('overrides'):
```
kdmccormick, line 55:
```suggestion
        if block_entry_point.group.endswith('.overrides'):
```

super-nit: we're appending '.overrides' so we may as well check for that same exact suffix

xblock/plugin.py
```diff
@@ -41,6 +41,32 @@ def default_select(identifier, all_entry_points):  # pylint: disable=inconsisten
...
+            block_entry_points.append(block_entry_point)
+
+    # Overrides get priority
+    if len(overrides) == 1:
+        return overrides[0]
+    elif len(overrides) > 1:
+        raise AmbiguousPluginError(all_entry_points)
+
+    # ... then default behavior
+    return default_select(identifier, block_entry_points)
```
kdmccormick, line 67: IMO a good defensive behavior would be to raise PluginMissingError in the case where an `xblock.v1.overrides` entry exists but the corresponding regular `xblock.v1` doesn't exist. You could achieve that here by performing default selection before checking the overrides list.

My thinking on this is that it would be weird to try to override something that doesn't exist in the first place, and the operator would probably want to be alerted to that weirdness. If I'm wrong, then we could always relax this assumption in the future without breaking anyone, whereas tightening the assumption in the future would be a breaking change.

xblock/plugin.py
```diff
@@ -98,9 +124,13 @@ def select(identifier, all_entry_points):
         if key not in PLUGIN_CACHE:
 
             if select is None:
-                select = default_select
+                select = select_with_overrides
```
kdmccormick, line 127: since the default behavior has changed, can you update the docstring?

CHANGELOG.rst
```diff
@@ -5,6 +5,12 @@ Change history for XBlock
 Unreleased
 ----------
 
+5.1.0 - 2024-08-07
```
kdmccormick, line 8: Thanks for the changelog entry. Could you update the docs too? Maybe just a quick "Replace a preinstalled XBlock with a custom implementation" guide under https://github.com/openedx/XBlock/tree/master/docs/xblock-tutorial/edx_platform (which [publishes here](https://docs.openedx.org/projects/xblock/en/latest/xblock-tutorial/edx_platform/index.html)) ?

The docs in that directory are admittedly a mess (although we [have a volunteer working on them](https://github.com/openedx/XBlock/issues/774) :), so don't feel like you need to write anything stellar.

xblock/plugin.py
kdmccormick: I like your instinct @farhan to avoid expanding the public API unnecessarily. Another approach would be call this new function `default_select`, since it is indeed the default selection method now. The old default select could become a private method, like `_default_select_no_override`

kdmccormick (changes requested):
Very nice implementation. Thanks so much for jumping in and making this happen. Just a few requests.

xblock/plugin.py
farhan: or we can remove `_default_select_no_override` as its not required anymore for simplicity.

CHANGELOG.rst
nsprenkle: @kdmccormick , will tackle in a follow-up PR.

nsprenkle:
@kdmccormick and @farhan , have updated to your suggestions. Ready for re-review.

CHANGELOG.rst
nsprenkle: JK, didn't realize this was the same repo. Will add it shortly.

nsprenkle: @kdmccormick , docs updated here: https://github.com/openedx/XBlock/pull/778/commits/5720d4a934e0cafb834d8c85a0b7641bb1f49811

docs/xblock-tutorial/edx_platform/overrides.rst
```diff
@@ -0,0 +1,75 @@
...
+
+        entry_points={
+            "xblock.v1": [
+                "tpp = xblocks.twitch_plays_pokemon::TwitchPlaysPokemonBlock"
+                # ...
+            ]
+        }
+    )
+
+If you then create your own Plugin App with a custom version of that XBlock...
```
kdmccormick, line 47:
```suggestion
If you then create your own Python package with a custom version of that XBlock...
```

any Python package will work, whether or not it uses the Open edX Django plugin system

docs/xblock-tutorial/edx_platform/overrides.rst
```diff
@@ -0,0 +1,75 @@
...
+        # ...
+        entry_points={
+            "xblock.v1.overrides": [
+                "tpp = your_plugin.twitch_plays_pokemon::YourTwitchPlaysPokemonBlock"
+                # ...
+            ],
+        }
+    )
+
+And install that app as a Django plugin, then your block should be loaded instead of the
```
kdmccormick, line 66:
```suggestion
And install that package into your virtual environment, then your block should be loaded instead of the
```

Django plugin mechanism isn't required here -- the block's package just needs to be installed

docs/xblock-tutorial/edx_platform/overrides.rst
```diff
@@ -0,0 +1,75 @@
...
+*******
+Example
+*******
+
+Imagine there is an XBlock installed ``edx-platform``:
+
+.. code:: python
+
+    # edx-platform/xblocks/twitch_plays_pokemon.py
+    class TwitchPlaysPokemonBlock(XBlock):
```
kdmccormick, line 32: sorry to be a grouch, but I'm not sure the whimsy meshes with the tone of our existing admin-facing docs 😛 

perhaps, to be maximially concrete, it would be helpful to reference a real edx-platform XBlock, like edx-platform/xmodule/poll_block.py ?

xblock/test/test_plugin.py
```diff
@@ -52,6 +56,29 @@ def boom(identifier, entry_points):
...
+    # Trying to load a block that is overridden, but ambigous, errors.
+    expected_msg = (
+        "Ambiguous entry points for overridden_block: "
+        "xblock.test.test_plugin.AmbiguousBlock1, "
+        "xblock.test.test_plugin.AmbiguousBlock2"
+    )
+    with pytest.raises(AmbiguousPluginError, match=expected_msg):
+        XBlock.load_class("overridden_block")
+
+
```
kdmccormick, line 81: could you add one more test case to ensure that PluginMissing is raised when there is an xblock.v1.overrride but no xblock.v1 ?

kdmccormick (approved):
A few more comments but nothing that needs re-review. I haven't manually tested, so I'm approving on the assumption that you have.

Thank a bunch @nsprenkle ! This will be awesome for site operators to have 🎸

xblock/plugin.py
```diff
@@ -75,12 +110,20 @@ def _load_class_entry_point(cls, entry_point):
...
-        the first one and log a warning.
+        By default, this returns the class mapped to `identifier` from entry_points
+        matching `{cls.entry_points}.overrides` or `{cls.entry_points}`, in that order.
+
+        If multiple classes are found for either `{cls.entry_points}.overrides` or
+        `{cls.entry_points}`, it will raise an `AmbiguousPluginError`.
 
-        If `default` is provided, return it if no entry_point matching
-        `identifier` is found. Otherwise, will raise a PluginMissingError
+        If no classes are found for `{cls.entry_points}`, it will raise a `PluginMissingError`.
```
farhan, line 119: Swap both lines as per order
```python
        If no classes are found for `{cls.entry_points}`, it will raise a `PluginMissingError`.


        If multiple classes are found for either `{cls.entry_points}.overrides` or
        `{cls.entry_points}`, it will raise an `AmbiguousPluginError`.
```

farhan (changes requested):

docs/xblock-tutorial/edx_platform/overrides.rst
```diff
@@ -0,0 +1,75 @@
...
+    # your_plugin/xblocks/video_block.py
+    class YourVideoBlock(XBlock):
+    ...
+
+    # your_plugin/setup.py
+    setup(
+        # ...
+        entry_points={
+            "xblock.v1.overrides": [
+                "tpp = your_plugin.xblocks.video_block::YourVideoBlock"
```
farhan, line 60: Should it not be
`"video = your_plugin.xblocks.video_block::YourVideoBlock"`

xblock/plugin.py
nsprenkle: Arguably, `load_class` doesn't even care about those details since those reside in the `select` function. I would prefer to delete them entirely than split hairs over the ordering.

kdmccormick: > Arguably, load_class doesn't even care about those details since those reside in the select function. 

+1, they are good as is

FWIW they are helpful lines to have in the docstring

xblock/plugin.py
```diff
@@ -41,6 +45,37 @@ def default_select(identifier, all_entry_points):  # pylint: disable=inconsisten
...
+
+    # Get the default entry point
+    default_plugin = _default_select_no_override(identifier, block_entry_points)
+
+    # If we have an unambiguous override, that gets priority. Otherwise, return default.
+    if len(overrides) == 1:
+        return overrides[0]
+    elif len(overrides) > 1:
+        raise AmbiguousPluginError(all_entry_points)
+    return default_plugin
```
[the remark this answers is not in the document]
kdmccormick: Regarding _default_select_no_override -- Yeah, I do not feel strongly since it is an internal implementation detail, but combining the functions does make it easier to follow IMO. If you are willing to make that change @nsprenkle, great; if not, I don't want to block merging on this concern.

Regarding the error message -- @farhan makes a very good point, we should make it clear whether there are multiple regular plugins vs. multiple overrides. As it is, the error cases would be indistinguishable to an operator. 

@nsprenkle , I don't have a preference between a new exception class vs. the same exception with a different message. If you do go with separate exceptions, please just make AmbiguousPluginOverrideError a subclass of AmbiguousPluginError so that existing `catch AmbiguousPluginError` statements will catch AmbiguousPluginOverrideError.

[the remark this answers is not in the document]
nsprenkle: @kdmccormick , added as a new Exception

kdmccormick (approved):

nsprenkle:
@kdmccormick / @farhan. Will merge / create a new release on Monday. Thanks, both!

farhan (approved):
@nsprenkle Nice work 🌟 
Squash merge it and release the new version

Issue #811, opened 2025-01-18
Poor performance when XBlock class is not found


ormsbee:
Checking the entrypoints for the XBlock class we want for a given tag name is relatively expensive. We mitigate this by caching all the XBlock classes when we load them. But we don't cache when we fail to find an XBlock class. So if a course has many instances of a type of XBlock that is not installed on the given instance, we're constantly scanning entrypoints. This really slows down things like the Studio course outline page.

The fix is to cache the fact that there is no matching XBlock class for this tag.

ormsbee:
On one of our sandbox instances, this accounts for ~85% of a five second load time for the Studio course index page.

PR #813, opened 2025-01-19
perf: cache missing XBlock mappings in load_class

ormsbee:
Checking the entrypoints for the XBlock class we want for a given tag name is relatively expensive. We mitigate this by caching the mapping of tags to classes in Plugin.load_class(). But before this commit, we weren't caching failed lookups, i.e. the situation where there is no XBlock that maps to a given tag. This meant that if a course had many instances of a type of XBlock that is not installed on the given server, we would be constantly scanning entrypoints to look for our non-existent XBlock subclass. This really slows down things like the Studio course outline page.

The fix here is to store failed lookups as None values in the PLUGIN_CACHE.

### DIFF
--- a/xblock/__init__.py
+++ b/xblock/__init__.py
@@ -2,4 +2,4 @@
 XBlock Courseware Components
 """
 
-__version__ = '5.1.0'
+__version__ = '5.1.1'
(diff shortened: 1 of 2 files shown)


ormsbee:
Fixes https://github.com/openedx/XBlock/issues/811

feanil (approved):

PR #815, opened 2025-01-28
fix: throw an error if the scope if None on scope_mappings [FC-0076]

rpenido:
## Description

While using the XBlock API, we may not have a `FieldSet` for a specific scope, which resulted in `AttributeError: 'NoneType' object has no attribute 'foo'` errors.

With this PR, the `InvalidScopeError` is now thrown in these cases.

## Testing instructions
- Edit a `Survey` block on the Library Authoring page and check the error below (before https://github.com/openedx/edx-platform/pull/36226)

```
cms-1          | 2025-01-28 19:36:18,402 ERROR 2163 [celery_utils.logged_task] [user 4] [ip 172.18.0.1] logged_task.py:48 - [9d7cda4b-afd2-404f-a39c-7a185c486907] failed due to Traceback (most recent call last):
cms-1          |   File "/openedx/venv/lib/python3.11/site-packages/celery/app/trace.py", line 453, in trace_task
cms-1          |     R = retval = fun(*args, **kwargs)
cms-1          |                  ^^^^^^^^^^^^^^^^^^^^
cms-1          |   File "/openedx/venv/lib/python3.11/site-packages/celery/app/autoretry.py", line 38, in run
cms-1          |     return task._orig_run(*args, **kwargs)
cms-1          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cms-1          |   File "/openedx/venv/lib/python3.11/site-packages/edx_django_utils/monitoring/internal/code_owner/utils.py", line 195, in new_function
cms-1          |     return wrapped_function(*args, **kwargs)
cms-1          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cms-1          |   File "/openedx/edx-platform/openedx/core/djangoapps/content/search/tasks.py", line 57, in upsert_library_block_index_doc
cms-1          |     api.upsert_library_block_index_doc(usage_key)
cms-1          |   File "/openedx/edx-platform/openedx/core/djangoapps/content/search/api.py", line 639, in upsert_library_block_index_doc
cms-1          |     searchable_doc_for_library_block(library_block_metadata)
cms-1          |   File "/openedx/edx-platform/openedx/core/djangoapps/content/search/documents.py", line 381, in searchable_doc_for_library_block
cms-1          |     block = xblock_api.load_block(xblock_metadata.usage_key, user=None)
cms-1          |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cms-1          |   File "/openedx/edx-platform/openedx/core/djangoapps/xblock/api.py", line 117, in load_block
cms-1          |     return runtime.get_block(usage_key, version=version)
cms-1          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cms-1          |   File "/openedx/edx-platform/openedx/core/djangoapps/xblock/runtime/learning_core_runtime.py", line 229, in get_block
cms-1          |     block.force_save_fields(block._get_fields_to_save())  # pylint: disable=protected-access
cms-1          |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cms-1          |   File "/mnt/xblock/xblock/core.py", line 531, in force_save_fields
cms-1          |     self._field_data.set_many(self, fields_to_save_json)
cms-1          |   File "/mnt/xblock/xblock/field_data.py", line 164, in set_many
cms-1          |     field_data.set_many(block, new_update_dict)
cms-1          |     ^^^^^^^^^^^^^^^^^^^
cms-1          | AttributeError: 'NoneType' object has no attribute 'set_many'
```
- Checkout the current PR and mount it on your tutor stack
- Check that the error throw now is the correct `InvalidScopeError`

___
Private ref: [FAL-4033](https://tasks.opencraft.com/browse/FAL-4033)

### DIFF
--- a/xblock/field_data.py
+++ b/xblock/field_data.py
@@ -145,10 +145,12 @@ class SplitFieldData(FieldData):
         """Return the field data for the field `name` on the :class:`~xblock.core.XBlock` `block`"""
         scope = block.fields[name].scope
 
-        if scope not in self._scope_mappings:
+        scope_mapping = self._scope_mappings.get(scope)
+
+        if scope_mapping is None:
             raise InvalidScopeError(scope)
 
-        return self._scope_mappings[scope]
+        return scope_mapping
 
     def get(self, block, name):
         return self._field_data(block, name).get(block, name)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,11 @@ Change history for XBlock
 Unreleased
 ----------
 
+5.1.2 - 2025-02-07
+------------------
+
+* throws InvalidScopeError if the scope is not defined
+
 5.1.0 - 2024-08-07
 ------------------
 
--- a/xblock/__init__.py
+++ b/xblock/__init__.py
@@ -2,4 +2,4 @@
 XBlock Courseware Components
 """
 
-__version__ = '5.1.1'
+__version__ = '5.1.2'
--- a/xblock/test/test_field_data.py
+++ b/xblock/test/test_field_data.py
@@ -40,6 +40,11 @@ class TestSplitFieldData:
             Scope.content: self.content,
             Scope.settings: self.settings
         })
+        self.split_empty = SplitFieldData({
+            Scope.content: self.content,
+            Scope.settings: self.settings,
+            Scope.user_state: None,
+        })
         self.runtime = TestRuntime(services={'field-data': self.split})
         self.block = TestingBlock(
             runtime=self.runtime,
@@ -75,6 +80,10 @@ class TestSplitFieldData:
     def test_invalid_scope(self):
         with pytest.raises(InvalidScopeError):
             self.split.get(self.block, 'user_state')
+
+    def test_empty_scope(self):
+        with pytest.raises(InvalidScopeError):
+            self.split_empty.get(self.block, 'user_state')
 
     def test_default(self):
         self.split.default(self.block, 'content')


xblock/core.py
```diff
@@ -687,7 +687,7 @@ class XBlock(Plugin, Blocklike, metaclass=_HasChildrenMetaclass):
 
     # These are dynamically managed by the awful hackery of _HasChildrenMetaclass.
     # We just declare their types here to make static analyzers happy.
-    # Note that children is only defined iff has_children is defined and True.
+    # Note that children is only defined if has_children is defined and True.
```
pomegranited, line 690: nit: this comment change can be reverted -- [iff](https://www.dictionary.com/browse/iff) is a shorthand term for "if and only if"

pomegranited (approved):
:+1: Fixes the issue, thanks @rpenido !

- [x] I tested this on my tutor dev stack by saving changes to the Survey & Poll XBlocks
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] Includes documentation - inline code comment explaining the change
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

xblock/core.py
rpenido: I didn't know that! Thanks @pomegranited!!
I dropped the commit.

rpenido:
Hi @pdpinch, @ormsbee and @felipemontoya!
Could any of you review this PR?
Thanks!

xblock/field_data.py
```diff
@@ -161,7 +161,8 @@ def set_many(self, block, update_dict):
         for key, value in update_dict.items():
             update_dicts[self._field_data(block, key)][key] = value
         for field_data, new_update_dict in update_dicts.items():
-            field_data.set_many(block, new_update_dict)
+            if field_data is not None:  # Ignore fields from scopes that are not loaded
```
ormsbee, line 164: Do you think we should log a warning in this case?

bradenmacdonald: Yeah, I don't think this error should be silently ignored at this level. In fact, I think raising an exception is the right behavior. If we choose to ignore unsupported scopes like `Scope.user_state`, we should be doing that explicitly by mapping that scope to a no-op.

> As the `LearningCoreXBlockRuntime` doesn't have the `Scope.user_state`, the get_block function throws an error.

I don't really understand this. [We _do_ support the `Scope.user_state` scope in Learning Core runtime](https://github.com/openedx/edx-platform/blob/c84af052517b46c95a6c0600fe9bbace920188f0/openedx/core/djangoapps/xblock/runtime/runtime.py#L385). Although I just noticed now that there is one circumstance where we don't: when `user is None`:

https://github.com/openedx/edx-platform/blob/c84af052517b46c95a6c0600fe9bbace920188f0/openedx/core/djangoapps/xblock/runtime/runtime.py#L353-L367

So: I suspect the root problem here is that **you're trying to make changes to `choices` when `user` is `None`**, which makes no sense. But if `user` is set, then something else weird is going on. In any case, I don't think ignoring it is the right fix.

bradenmacdonald: Ah, I see now we're not really modifying any fields, just marking them as loaded.

One fix could be to change [this line](https://github.com/openedx/edx-platform/blob/c84af052517b46c95a6c0600fe9bbace920188f0/openedx/core/djangoapps/xblock/runtime/learning_core_runtime.py#L225) to omit any fields from user-specific scopes if `user=None`, but that seems fragile.

A simpler approach is probably to change [this line](https://github.com/openedx/edx-platform/blob/c84af052517b46c95a6c0600fe9bbace920188f0/openedx/core/djangoapps/xblock/runtime/runtime.py#L355) so that instead of setting `student_data_store = None` , we set `student_data_store` to [`DictFieldData`](https://github.com/openedx/XBlock/blob/03251b23500183f2b6ee32e056c0447cf812753a/xblock/field_data.py#L104) to explicitly ignore any writes there. It could be a subclass that logs warnings, but the true root cause here is that XBlock Field Data API requires calling `set` when initializing the block, so we'd be issuing a pointless warning because nobody's actually trying to write to that field; the field simply exists and we're loading the initial value.

rpenido: The underlying issue is that the `choices` field is marked as `dirty` on load. I'll dig a bit deeper and try to fix that.

In retrospect, I think we should throw an exception if the scope is not set on `scope_mappings`. We have other blocks that throw the inconclusive `AttributeError: 'NoneType' object has no attribute 'set_many'` (i.e. `problem-builder` because of `Scope.children` being `None`).

ormsbee: Yeah, XBlock fields are marked dirty on read if their types are mutable, i.e. lists, sets, dicts. I have a [rant blog post](https://medium.com/@dormsbee/xblock-lessons-costly-conveniences-38ff8e7de6f1) that talks about this.

rpenido: Thanks @ormsbee!
I just found that. :disappointed:

Not sure the best way to go from here:
1. Clean `user_state_*` fields from `get_fields_to_save()` on load
2. Use the 'DictFieldData` as `student_data_store`
3. ????

rpenido: Hi @bradenmacdonald! Do you think we should go on the `DictFieldData` route?

bradenmacdonald: @rpenido I'm leaning that way. I think it's a relatively simple fix for the problem you're seeing.

rpenido: Thanks @bradenmacdonald! PR created: https://github.com/openedx/edx-platform/pull/36226
I changed this current PR to throw `InvalidScopeError` instead of `AttributeError` if the scope is not set (it will not happen anymore for the block load, but it can occur in other cases).

xblock/field_data.py
```diff
@@ -148,7 +148,12 @@ def _field_data(self, block, name):
         if scope not in self._scope_mappings:
             raise InvalidScopeError(scope)
```
bradenmacdonald, line 149: Isn't this line trying to do something similar? Maybe you can consolidate these two conditions using `self._scope_mappings.get(...)`

bradenmacdonald (changes requested):
Thanks, seems good, just a minor request.

xblock/field_data.py
rpenido: Done: e048432336cad94554a4096edc6f42a23d8d7969

bradenmacdonald (approved):

bradenmacdonald:
Approved! Can you try squashing and/or rebasing this? I'm not sure why the docs check hasn't passed, but it's blocking me from merging.

bradenmacdonald:
Also, please include a version bump if you'd like to. But it's also fine to just leave this "unreleased" and wait until it gets combined with some other changes. With the other fix we have in place, I don't think this PR is urgent anymore.

rpenido:
Squased and Rebased!
But the readthedocs pending request persists. :disappointed: 
Also, I bumped the package version.

bradenmacdonald:
Closed and re-opened to see if that will fix the docs check.

rpenido:
Thanks for merging it @bradenmacdonald!
I think it got auto-tagged, but could you create a release?

bradenmacdonald:
@rpenido Ah, I tagged it but I guess that wasn't enough. I've done a release now too.

Issue #823, opened 2025-03-12
Add logic to parse XBlock's XML with the pointer tag correctly.


ttqureshi:
Currently `parse_xml` function in [xblock/core.py](https://github.com/openedx/XBlock/blob/master/xblock/core.py) doesn't support the implementation to parse the pointer tags within an XBlock's XML file. It instantiates the XBlock without picking fields data (and other metadata, if any) if there is a pointer tag.

**Example XML with pointer tag**
```
<vertical display_name="LTI">
  <lti url_name="lti"/>
</vertical>
```

## Why need it?
We are extracting BuiltIn XBlocks from edx-platform, in order for the extracted XBlocks to work properly we want this pointer tag logic to be implemented in `xblock/core.py`. This will prevent breaking any functionality, ensuring that XBlock's definition is loaded correctly.

Here's the reference from Demo course, which was exported from Studio as recently as Redwood:
- within a vertical [LTI block pointer tag](https://github.com/openedx/openedx-demo-course/blob/master/demo-course/course/vertical/996f72a9a8b9429d8f887bd91c2e64cc.xml#L6)
- pointing to the LTI Block definition [here](https://github.com/openedx/openedx-demo-course/blob/master/demo-course/course/lti/8e71d278e94f4913bc564cc6141fde53.xml).

ttqureshi:
FYI @kdmccormick

cc: @farhan @salman2013 @irtazaakram

kdmccormick:
TY. Out on vacation currently but I will add some notes when I'm back on Friday.

kdmccormick:
I have created this ticket for the overall bug investigation:

* https://github.com/openedx/edx-platform/issues/36390

We can keep this issue (xblock/823) focused on fixing the immediate bug, and use edx-platform/36390 for the big-picture investigation of what happened.

kdmccormick:
There is one big-picture thing we need to worry about for this ticket:

In order to keep Teak course exports compatible with the broken releases (which are most likely Redwood and Sumac), _we should continue to export the currently-external blocks using inline syntax_, even though we will restore the ability to import them with pointer syntax. I think we could do this by adding a new class variable to XBlock called `export_with_pointer_tag`, defaulting to False. The currently-built-in blocks and their extracted counterparts can override this variable to True. The export process can look at the variable to determine whether to export that block using inline or pointer syntax.

[...]

PR #830, opened 2025-03-24
Adds pointer tag parsing implementation in `parse_xml`

ttqureshi:
- resolves: https://github.com/openedx/XBlock/issues/823


ttqureshi:
This PR is ready for review.
cc: @kdmccormick @feanil

farhan (changes requested):

xblock/core.py
```diff
@@ -32,12 +32,85 @@
...
 
+XML_PARSER = etree.XMLParser(dtd_validation=False, load_dtd=False, remove_blank_text=True, encoding='utf-8')
+
 # __all__ controls what classes end up in the docs.
 __all__ = ['XBlock', 'XBlockAside']
 
 UNSET = object()
 
 
+def name_to_pathname(name):
```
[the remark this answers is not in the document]
ttqureshi: > Have you copied this methods/logic from somewhere else?

Yep,
I've taken it from `XmlMixin`'s [pointer tags parsing](https://github.com/ttqureshi/edx-platform/blob/9c4c0495f4aa9bd4a1c91e6d4d65ed47b32a26da/xmodule/xml_block.py#L30) implementation.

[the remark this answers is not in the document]
ttqureshi: > I think it will be a good idea to refactor these util methods into a separate util class to keep this core class minimal and should have only core functionality.

ON IT!

farhan (changes requested):

xblock/core.py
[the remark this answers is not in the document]
ttqureshi: ✔️

requirements/dev.txt
```diff
@@ -112,7 +112,7 @@ fs-s3fs==1.1.1
     # via
     #   -r requirements/test.txt
     #   openedx-django-pyfs
-hypothesis==6.130.8
+hypothesis==6.130.10
```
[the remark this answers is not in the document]
ttqureshi: resolved ✔️

xblock/core.py
```diff
@@ -746,6 +748,11 @@ def parse_xml(cls, node, runtime, keys):
             keys (:class:`.ScopeIds`): The keys identifying where this block
                 will store its data.
         """
+        if is_pointer_tag(node):
+            # new style:
```
farhan, line 752: I think we can remove the comment and add better docs on load_definition_xml method

xblock/utils/helpers.py
```diff
@@ -23,3 +27,74 @@ def child_isinstance(block, child_id, block_class_or_mixin):
...
+    Loads definition_xml stored in a dedicated file
+    """
+    url_name = node.get('url_name')
+    filepath = format_filepath(node.tag, name_to_pathname(url_name))
+    definition_xml = load_file(filepath, runtime.resources_fs, def_id)
+    return definition_xml, filepath
+
+
+def format_filepath(category, name):
+    return f'{category}/{name}.xml'
```
farhan, line 75: Add docs

xblock/utils/helpers.py
```diff
@@ -23,3 +27,74 @@ def child_isinstance(block, child_id, block_class_or_mixin):
...
+    actual_attr = set(xml_obj.attrib.keys())
+
+    has_text = xml_obj.text is not None and len(xml_obj.text.strip()) > 0
+
+    return len(xml_obj) == 0 and actual_attr == expected_attr and not has_text
+
+
+def load_definition_xml(node, runtime, def_id):
+    """
+    Loads definition_xml stored in a dedicated file
```
farhan, line 66: Following sounds better
```python
    """
        Parses and loads an XML definition file based on a given node, runtime
        environment, and definition ID.

        Arguments:
        node: XML element containing attributes for definition loading.
        runtime: The runtime environment that provides resource access.
        def_id: Unique identifier for the definition being loaded.

        Returns:
        tuple: A tuple containing the loaded XML definition and the
        corresponding file path.
    """
```

xblock/test/utils/test_helpers.py
```diff
@@ -78,3 +88,59 @@ def test_child_isinstance_descendants(self):
...
+        definition_xml, filepath = load_definition_xml(node, runtime, def_id)
+        self.assertEqual(filepath, "course/test_url.xml")
+        self.assertEqual(definition_xml, "<mock_xml />")
+        mock_load_file.assert_called_once()
+
+    def test_format_filepath(self):
+        self.assertEqual(format_filepath("course", "test_url"), "course/test_url.xml")
+
+    @patch("xblock.utils.helpers.etree.parse")
+    def test_file_to_xml(self, mock_etree_parse):
```
farhan, line 141: Following seems more reasonable test case to me
```
class TestFileToXml(TestCase):
    def test_valid_xml_file(self):
        file_mock = MagicMock()
        file_mock.read.return_value = b"<root><child>Value</child></root>"
        result = file_to_xml(file_mock)
        self.assertEqual(result.tag, 'root')
        self.assertEqual(result[0].tag, 'child')
        self.assertEqual(result[0].text, 'Value')
```

farhan (approved):

farhan:
I didn't test this PR manually, I am trusting Tayyab good domain knowledge and testing on it

ttqureshi:
@kdmccormick 
A gentle reminder, your review is awaited for this PR. It's a blocker for [LTIBlock Test PR](https://github.com/openedx/edx-platform/pull/36020).

kdmccormick:
Yep, this is in my queue Tayyab, sorry for my slowness.

ttqureshi:
> Yep, this is in my queue Tayyab, sorry for my slowness.

No worries

xblock/core.py
```diff
@@ -23,6 +23,7 @@
 from xblock.fields import Field, List, Reference, ReferenceList, Scope, String
 from xblock.internal import class_lazy
 from xblock.plugin import Plugin
+from xblock.utils.helpers import is_pointer_tag, load_definition_xml
```
kdmccormick, line 26: Unlike in other repos, the `utils`, package in xblock is not meant to hold core helper methods. It actually holds _edx-platform developer utilities_ (e.g. StudioEditableMixin)-- the name is historical. `xblock` doesn't ever import from `xblock.utils`, and it would help the code stay simple if we kept it that way.

Could you move the new helpers to a new file `xblock/xml.py` ?

xblock/utils/helpers.py
```diff
@@ -23,3 +27,87 @@ def child_isinstance(block, child_id, block_class_or_mixin):
...
+    def_id: Unique identifier for the definition being loaded.
+
+    Returns:
+    tuple: A tuple containing the loaded XML definition and the
+    corresponding file path.
+    """
+    url_name = node.get('url_name')
+    filepath = format_filepath(node.tag, name_to_pathname(url_name))
+    definition_xml = load_file(filepath, runtime.resources_fs, def_id)
+    return definition_xml, filepath
```
kdmccormick, line 81: Would this work in xblock-sdk as well, or is this logic specific to edx-platform?

If it is specific to edx-platform, then we want want to go through a Runtime method instead, so that edx-platform and xblock-sdk can  implement it differently.

xblock/utils/helpers.py
```diff
@@ -23,3 +27,87 @@ def child_isinstance(block, child_id, block_class_or_mixin):
...
+    """
+    Open the specified file in fs, and call cls.file_to_xml on it,
+    returning the lxml object.
+
+    Add details and reraise on error.
+    """
+    try:
+        with fs.open(filepath) as xml_file:
+            return file_to_xml(xml_file)
+    except Exception as err:  # lint-amnesty
```
kdmccormick, line 101:
```suggestion
    except Exception as err:
```
`lint-amnesty` is supposed to indicate pylint violations that were ignored, in-bulk, at one specific time several years ago. You should never add the token `lint-amnesty` to any new code.

kdmccormick (changes requested):

xblock/utils/helpers.py
ttqureshi: I have added the `load_definition_xml` function to the Runtime, which xblock-sdk can override. 
I have not made it abstract method because this function has been taken form the `XmlMixin` in the edx-platform.
In order to make it abstract and raise the `NotImplementedError` in the `Runtime` class, we'll have to make changes in the edx-platform correspondingly that are extending the `Runtime` .

ttqureshi: > Would this work in xblock-sdk as well, or is this logic specific to edx-platform?
> 
> If it is specific to edx-platform, then we want want to go through a Runtime method instead, so that edx-platform and xblock-sdk can implement it differently.

@kdmccormick 
I have added the `load_definition_xml` function to the Runtime, which xblock-sdk can override. 
I have not made it abstract method because this function has been taken form the `XmlMixin` in the edx-platform.
In order to make it abstract and raise the `NotImplementedError` in the `Runtime` class, we'll have to make changes in the edx-platform correspondingly that are extending the `Runtime` .

bradenmacdonald:
I'm not so sure about this approach and wrote up some questions [here](https://github.com/openedx/XBlock/issues/823#issuecomment-2881707131) on the related issue. I don't want to unnecessarily hold things up so don't take this as a definite blocker, but I do want to say that:

In general, the way that `parse_xml` (and the related methods like `parse_xml_new_runtime`, `definition_from_xml`, `load_definition_xml`, etc.) works in edx-platform is very messy and confusing. A big part of the problem is the inconsistent use of `XmlMixin` and the complex method resolution order of all our mixins in general, but the main problem is just the stupid API contract of `parse_xml`, `add_node_as_child`, etc. But given that we can't really change the latter anytime soon, I would strongly prefer that we find ways to make the platform's XML parsing simpler and more consistent, rather than adding more divergence and complexity to the core XBlock parsing APIs.

kdmccormick (changes requested):
We are discussing alternative implementations on https://github.com/openedx/XBlock/issues/823#issuecomment-2936185369

ttqureshi:
Closing this PR in favor of this [PR](https://github.com/openedx/edx-platform/pull/37133) in the edx-platform.

PR #906, opened 2026-04-14
feat: move commonly used attributes from XModuleMixin to XBlock

irtazaakram:
This pull request is related to https://github.com/openedx/openedx-platform/pull/38271

It moves a set of commonly used attributes out of the legacy `XModuleMixin` into `XBlock`, since they are not specific to `XModule` behavior and are widely relied upon across most XBlocks. 

The update also includes corresponding tests (generated with assistance from Claude Code), which have been carefully reviewed and refined.

### DIFF
--- a/xblock/core.py
+++ b/xblock/core.py
@@ -970,6 +970,7 @@ class XBlock(Plugin, Blocklike, metaclass=_HasChildrenMetaclass):
         """
         return getattr(self, "icon_class", "other")
 
+    @property
     def display_name_with_default(self):
         """
         Return a display name for this block.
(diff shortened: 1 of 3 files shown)


xblock/core.py
```diff
@@ -938,6 +964,35 @@ def has_support(self, view, functionality):
...
+        Return a display name for this block.
+
+        Uses ``display_name`` if it is set and not None. Otherwise, falls back
+        to a name derived from the block's ``usage_key.block_id`` by replacing
+        underscores with spaces.
+
+        This method avoids direct attribute access and uses ``getattr`` to safely
+        handle cases where attributes like ``display_name`` or ``usage_key`` may
+        not be present.
+
```
kdmccormick, line 985: `usage_key` is always present. Anyway, this is an implementation detail, so it is better to leave it out of the docstring.
```suggestion
```

kdmccormick (approved):
Just one request, good to merge otherwise.

PR #910, opened 2026-05-08
docs: Add internal ref to XBlock Intro doc

sarina:
Small change - so I can cross-reference docs


sarina:
Ahhh I missed another doc, this PR is unnecessary.

PR #934, opened 2026-06-29
fix: make Mixologist async-safe, deprecate ObjectAggregator

### LINKED ISSUE #918: Investigate and possibly replace threading RLock in ObjectAggregator
I'm not sure if this is async safe, or if everything happens up front at startup anyway:

https://github.com/openedx/XBlock/blob/77bfe4d7b72bcd09565c41595f6b8fa20c6d7087/xblock/runtime.py#L1189-L1226

We need to make it async-safe for ASGI style deployments.

Open question: Is this used anywhere at all? It looks like it might no longer be called anywhere other than its own tests.

irfanuddinahmad:
## Summary

- Removes `threading.RLock` (`_CLASS_CACHE_LOCK`) from `Mixologist`'s class cache, making `mix()` safe for ASGI/async deployments. `dict.setdefault()` atomicity (via CPython's GIL) provides equivalent thread safety without ever blocking the event loop.
- Deprecates `ObjectAggregator` with a `DeprecationWarning` — it has no production callers anywhere in XBlock or edx-platform and will be removed in a future major release.

## Why the lock had to go

In ASGI (single event-loop thread), calling `with threading.RLock():` stalls the **entire event loop** if another operation holds the lock, blocking every concurrent request. This is one of the threading hazards catalogued in the parent epic.

Two reasons no replacement lock is needed:

1. **asyncio is cooperative** — `mix()` has zero `await` points, so the scheduler can never interleave two calls on the same thread.
2. **CPython's GIL makes `dict.setdefault()` atomic** — under true multi-threading, two racing threads may both construct the `type(...)` object, but `setdefault` guarantees only one is stored. The other is discarded. This is a benign, one-time cost.

## Test plan

- [ ] All existing `TestMixologist` tests pass unchanged (caching and mixin behaviour is identical)
- [ ] `TestObjectAggregator` tests pass and assert the new `DeprecationWarning` fires on instantiation
- [ ] `grep threading xblock/runtime.py` returns nothing

Closes #918
Part of openedx/openedx-platform#38680

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/xblock/runtime.py
+++ b/xblock/runtime.py
@@ -11,7 +11,6 @@ import itertools
 import json
 import logging
 import re
-import threading
 import warnings
 
 from lxml import etree
@@ -1195,6 +1194,11 @@ class ObjectAggregator:
     """
 
     def __init__(self, *objects):
+        warnings.warn(
+            "ObjectAggregator is deprecated and will be removed in a future release.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         self.__dict__['_objects'] = objects
 
     def _object_with_attr(self, name):
@@ -1221,9 +1225,11 @@ class ObjectAggregator:
         delattr(self._object_with_attr(name), name)
 
 
-# Cache of Mixologist generated classes
+# Cache of Mixologist generated classes.
+# dict.setdefault() is atomic under CPython's GIL, so no lock is needed here.
+# This is also safe in asyncio contexts: mix() has no await points, so the
+# event loop cannot interleave two calls to it within the same thread.
 _CLASS_CACHE = {}
-_CLASS_CACHE_LOCK = threading.RLock()
 
 
 class Mixologist:
@@ -1277,18 +1283,16 @@ class Mixologist:
         mixin_key = (base_class, mixins)
 
         if mixin_key not in _CLASS_CACHE:
-            # Only lock if we're about to make a new class
-            with _CLASS_CACHE_LOCK:
-                # Use setdefault so that if someone else has already
-                # created a class before we got the lock, we don't
-                # overwrite it
-                return _CLASS_CACHE.setdefault(mixin_key, type(
-                    base_class.__name__ + 'WithMixins',  # type() requires native str
-                    (base_class,) + mixins,
-                    {'unmixed_class': base_class}
-                ))
-        else:
-            return _CLASS_CACHE[mixin_key]
+            new_class = type(
+                base_class.__name__ + 'WithMixins',  # type() requires native str
+                (base_class,) + mixins,
+                {'unmixed_class': base_class}
+            )
+            # setdefault is atomic under CPython's GIL: if two threads race
+            # here, one wins and the other's new_class is silently discarded.
+            _CLASS_CACHE.setdefault(mixin_key, new_class)
+
+        return _CLASS_CACHE[mixin_key]
 
 
 class RegexLexer:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,13 @@ Change history for XBlock
 Unreleased
 ----------
 
+* Removed ``threading.RLock`` from ``Mixologist``'s class cache to make it safe
+  for ASGI/async deployments. ``dict.setdefault()`` atomicity (via CPython's GIL)
+  provides equivalent thread safety without blocking the event loop.
+  Relates to `openedx/openedx-platform#38680 <https://github.com/openedx/openedx-platform/issues/38680>`_.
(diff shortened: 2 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @irfanuddinahmad!


This repository is currently maintained by `@openedx/axim-engineering`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

farhan (approved):

farhan:
@ormsbee PR looks good to me. It would be great if you could take a look

ormsbee (approved):
This looks fine to me as well, though I'm not sure if @kdmccormick wants to look at it.

This all happens at startup initialization before we start serving requests anyway, right? We're not actually mixing classes together during the request/response lifecycle, are we?

xblock/runtime.py
```diff
@@ -1277,18 +1283,16 @@ def mix(self, cls):
...
-                # created a class before we got the lock, we don't
-                # overwrite it
-                return _CLASS_CACHE.setdefault(mixin_key, type(
-                    base_class.__name__ + 'WithMixins',  # type() requires native str
-                    (base_class,) + mixins,
-                    {'unmixed_class': base_class}
-                ))
-        else:
-            return _CLASS_CACHE[mixin_key]
+            new_class = type(
```
kdmccormick, line 1286: Thank you both for reviewing.

@ormsbee , blocks are actually mixed on-demand, not at startup. If you throw a breakpoint here and start up a fresh CMS, it won't be hit until you load up a course, at which point you'll see it hit as part of the request handler once per loaded block type. Does that change your review at all?

ormsbee: It doesn't change my review, but it just seems like something that should happen at one time and place during startup (triggered from an AppConfig somewhere), unless there's a really compelling reason to delay it.

ormsbee: (With the justification that the Mixologist engages in weird magic, and weird magic should ideally happen only once and in obvious place and time.) To be clear, I'm not asking for this PR to change this behavior, especially since that's likely to be risky.

farhan: @ormsbee @kdmccormick So we are good to merge this PR, right?

